PR #20, opened 2015-09-22
travis testing against 0.12, 2, 4

ofrobots:
Setting up travis. @matthewloring PTAL.


ofrobots:
The failure against Node 4.x will be fixed by #21.

matthewloring:
LGTM once #21 is landed (just to confirm the fix)

PR #22, opened 2015-09-22
fix problem with server breakpoint deletion

ofrobots:
We were failing to notice when the last server breakpoint got deleted. As a
result we were keeping the V8 breakpoint, and the event listener active forever.
This was a performance problem.

I improved some log messages at the same time, while I was debugging.

@matthewloring PTAL.

### DIFF
--- a/lib/debuglet.js
+++ b/lib/debuglet.js
@@ -202,15 +202,13 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
           return;
 
         case 409: // Timeout on a hanging GET.
-          that.logger_.info('\t409 Hanging GET completed, no change.');
+          that.logger_.info('\t409 Long poll completed.');
           that.scheduleBreakpointFetch_(0/*immediately*/);
           return;
 
         default:
           that.logger_.info('\t' + response.statusCode + ' completed.');
-          if (body.breakpoints) {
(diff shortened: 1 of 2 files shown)


lib/debuglet.js
```diff
@@ -229,9 +227,11 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
  */
 Debuglet.prototype.updateActiveBreakpoints_ = function(breakpoints) {
   var that = this;
-
   var updatedBreakpointMap = this.convertBreakpointListToMap_(breakpoints);
-  that.logger_.breakpoints(Logger.INFO, 'Server breakpoints:', updatedBreakpointMap);
+
+  if (breakpoints.length) {
+    that.logger_.breakpoints(Logger.INFO, 'Server breakpoints:', updatedBreakpointMap);
```
matthewloring: At the INFO level, wouldn't it still be informative to know that there are no breakpoints on the server?

ofrobots: I'm preserving behavior with this PR. We weren't printing this before either. Also, the lack of output is pretty good signal at the INFO level. Printing the full response from the server would be more appropriate at DEBUG level.

matthewloring:
LGTM.

PR #25, opened 2015-09-25
Add coverage and dependency tracking

matthewloring:
This patch adds version number, coverage, and
dependency tracking badges to the readme. It
also introduces scripts for managing tests/
coverage/versioning and removes the no longer
used prepush hooks.

Fixes #5
Fixes #6

PR #26, opened 2015-09-25
Use env vars over config if present

### LINKED ISSUE #23: Numeric env vars are ignored if defined as 0
logging level is the current example of this

matthewloring:
Fixes #23

### DIFF
--- a/config.js
+++ b/config.js
@@ -17,11 +17,11 @@
 
 // Default configuration
 module.exports = {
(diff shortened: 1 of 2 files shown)


matthewloring:
@ofrobots PTAL.

matthewloring:
Tests for this require stand alone infrastructure which depends on #25. It's up next.

ofrobots:
LGTM.

PR #29, opened 2015-09-25
report error on breakpoint with invalid line num

ofrobots:
@matthewloring PTAL.

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -38,7 +38,8 @@
   SOURCE_MAP_READ_ERROR: 'The source map could not be read or was incorrectly formatted',
   V8_BREAKPOINT_MISSING: 'Internal error: V8 breakpoint missing',
   V8_BREAKPOINT_DISABLED: 'Internal error: V8 breakpoint externally disabled',
-  CAPTURE_BREAKPOINT_DATA: 'Error trying to capture breakpoint data: '
+  CAPTURE_BREAKPOINT_DATA: 'Error trying to capture breakpoint data: ',
+  INVALID_LINE_NUMBER: 'Breakpoint line number is invalid'
 };
 
 /** @const */ var MODULE_WRAP_PREFIX_LENGTH = require('module').wrap('☃')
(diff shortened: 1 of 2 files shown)


lib/v8debugapi.js
```diff
@@ -277,6 +280,12 @@ module.exports.create = function(logger_, config_) {
     // TODO: need test coverage for the above
     // TODO: work with debug server to come up with a more robust solution
 
+    if (breakpoint.location.line >= matchingScript.line_ends.length) {
+      return setErrorStatusAndCallback(cb, breakpoint,
+        StatusMessage.BREAKPOINT_SOURCE_LOCATION,
+        messages.INVALID_LINE_NUMBER);
```
matthewloring: Nit: indentation

matthewloring:
LGTM after the nit.

lib/v8debugapi.js
ofrobots: It is consistent with the other occurrences of setErrorStatusAndCallback in the file. Also, we need to pick a style-guide to refer to for this; and somehow enforce as well. What style guide should we use?

matthewloring: Ahh, I see that now. [The Google one](https://google.github.io/styleguide/javascriptguide.xml) seems like a good place to start but I am not yet familiar with it.

matthewloring: This patch appears to just check for scripts loaded by v8. We will have to make sure it applies generally when we move away from using v8 to find scripts.

ofrobots: Yes. The test must continue to be functional across any changes to how we figure out the the list of scripts. That's why this test needs to be added before we do any large refactoring of the script validation code.

ofrobots: The [style guide](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting) seems the allow the indentation I used, with the caveat that it prefers indent = 4 spaces while our code is indent = 2.

matthewloring: Yup, it looks weird to me on the github diff but it's not in violation of the guide.

matthewloring:
LGTM.

ofrobots:
Landed in https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/commit/fbac50d6570a1eb43ed0b72a9f4676e8d09a783f.

PR #30, opened 2015-09-29
compute a file index

ofrobots:
Start computing a static list of JS files

- Renamed hasher.js to scanner.js as now it can return both the list of JS files in the subtree as well as the hash-code.
- This paves the way to us using the static list of files to validate the break point requests against rather than the dynamic list of files loaded in V8.


ofrobots:
@matthewloring PTAL.

lib/scanner.js
```diff
@@ -0,0 +1,148 @@
...
+function computeHash(fileList, callback) {
+  var pending = fileList.length;
+  // return a valid, if fake, result when there are no js files to hash.
+  if (pending === 0) {
+    callback(null, 'EMPTY-no-js-files');
+    return;
+  }
+
+  var hashes = [];
+  fileList.forEach(function(filename) {
```
matthewloring: The `forEach` callback can take an index as its second argument. Using this will remove the need for pending.

matthewloring:
LGTM

PR #31, opened 2015-10-06
Resolve breakpoint target files statically

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/debuglet.js
```diff
@@ -83,42 +84,44 @@ function Debuglet(config, logger) {
...
-      that.logger_.error('Unable to compute unique id of the application', err);
-      if (opt_callback) {
-        opt_callback(err);
-      }
-      return;
-    }
-    that.logger_.info('Unique ID for this Application: ' + uid);
-
-    that.debugletApi_.init(uid, function(err) {
+  scanner.scan(that.config_.workingDirectory, function(err, hash, fileStats) {
```
ofrobots: Previous we would do a scan only if `workingDirectory` contained a `package.json`. I think this is important. This prevents us from doing a large scan of the whole file-system accidentally if the node process is started in a weird location. Could you bring back that behaviour?

ofrobots: Maybe you can bring that logic into this file and refuse to enable the debulet unless we can find a package.json in `config.workingDirectory`.

matthewloring: This logic has been moved and the uid file has been deleted.

matthewloring:
@ofrobots Ready for another look

test/test-uid.js
```diff
@@ -1,51 +0,0 @@
-/**
```
ofrobots: Can you keep these tests, perhaps move them to test-agent.js?

ofrobots: err... test-debuglet.js

matthewloring: Yup, they've been reintroduced.

ofrobots:
LGTM.

PR #32, opened 2015-10-29
env. var. GCLOUD_DEBUG_DISABLE doesn't have an effect

ofrobots:
GCLOUD_DEBUG_DISABLE should disable the agent.


ofrobots:
@matthewloring PTAL.

matthewloring:
LGTM.

PR #33, opened 2015-10-29
env. var. GCLOUD_DEBUG_LOGLEVEL doesn't have an effect

ofrobots:


test/standalone/test-env-log-level.js
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var LOGLEVEL = 4;
+process.env.GCLOUD_DEBUG_LOGLEVEL = LOGLEVEL;
```
matthewloring: I appreciate not hard coding constants but this feels a little silly as its only use is here.

test/standalone/test-env-log-level.js
```diff
@@ -0,0 +1,56 @@
...
+
+    var DebugletApi = proxyquire('../..', {
+      '@google/cloud-diagnostics-common': {
+        logger: logger,
+        utils: common.utils
+      }
+    });
+
+    setTimeout(function() {
+      assert(count > 0, 'Should have logged something');
```
matthewloring: Is there any way we can make a stronger assertion here? Do we know intended counts for any of the logs reported.

test/standalone/test-env-log-level.js
```diff
@@ -0,0 +1,56 @@
...
+var assert = require('assert');
+var proxyquire = require('proxyquire');
+var common = require('@google/cloud-diagnostics-common');
+
+// Mock the logger.
+var count = 0;
+var logger = {
+  create: function() {
+    return {
+      error: function() { count++; },
```
matthewloring: It would be nice to track these different log types separately and make sure we get the expected counts as opposed to one counter.

matthewloring:
@ofrobots Looks like JSHint has some semicolon related complaints. (and other style nits)

test/standalone/test-env-log-level.js
ofrobots: Agree. 'tis left over from an earlier version of the patch.

test/standalone/test-env-log-level.js
ofrobots: `logger` is harder to mock without duplication of logic. Ideally I would have patched `log()`, but that is closure private, and I can't get my fingers onto it. Let me see if there is simple change that gets me a stronger assertion.

test/standalone/test-env-log-level.js
ofrobots: done.

test/standalone/test-env-log-level.js
ofrobots: done.

test/standalone/test-env-log-level.js
ofrobots: Sufficiently addressed in the latest version of the test.

test/standalone/test-env-log-level.js
```diff
@@ -0,0 +1,44 @@
...
+    var agent = require('../../');
+    var logger = agent.private_.logger_;
+    var STRING1 = 'jjjjjjjjjjjjjjjjjfjfjfjf';
+    var STRING2 = 'kkkkkkkfkfkfkfkfkkffkkkk';
+
+    var buffer = [];
+    var oldLog = console.log;
+
+    console.log = function () {
+      buffer = buffer.concat([].slice.call(arguments));
```
matthewloring: Nit, I prefer `buffer.push.apply(buffer, arguments)` but it's not a big deal since this won't grow.

matthewloring:
LGTM.

PR #34, opened 2015-10-31
Do not modify protobuf breakpoint message

ofrobots:
The API rejects breakpoint updates if sees any unknown property on the
breakpoint object. It is not okay to add additional properties on breakpoint.

@matthewloring PTAL.


lib/v8debugapi.js
```diff
@@ -125,7 +125,7 @@ module.exports.create = function(logger_, config_, fileStats_) {
           source: scriptPath.indexOf('/') === -1 ? scriptPath
             : scriptPath.split('/').pop(),
           line: breakpoint.location.line,
-          column: 0
+          column: 0 // unlike lines, columns are 0-indexed for source-map ಠ_ಠ.
```
matthewloring: ಠ_ಠ

matthewloring:
LGTM

PR #36, opened 2015-11-04
Change the debuglet api to make it easier to test

ofrobots:
@matthewloring ptal.


index.js
```diff
@@ -33,7 +33,8 @@ if (process.env.hasOwnProperty('GCLOUD_DEBUG_DISABLE')) {
 }
 
 if (config.enabled) {
-  agent.start(config, logger.create(config.logLevel, '@google/cloud-debug'),
-      module.exports);
+  var debuglet = new Debuglet(
+    config, logger.create(config.logLevel, '@google/cloud-debug'));
+  debuglet.start();
```
matthewloring: Is there a reason we are not interested in errors coming from this call?

matthewloring:
LGTM but I agree that the callback to start should be eliminated when possible.

ofrobots:
There have been enough changes to warrant another look.

lib/debuglet.js
```diff
@@ -31,22 +34,14 @@ var NODE_VERSION_MESSAGE = 'V8 Debug API not available – make sure you are ' +
...
-    exports.private_ = agent;
-  }
-};
+module.exports = Debuglet;
 
 /**
  * @param {Object} config The option parameters for the Debuglet.
+ * @event 'error'
+ * @event 'started'
+ * @event 'registered'
```
matthewloring: Does registered have the connotation of "successfully initialized"? Could we document what an event of this type means.

matthewloring: Ah, it looks like `started` is the initialized event. Could we document what registered means.

matthewloring:
LGTM with small nit. I like the event emitter structure a lot more!

lib/debuglet.js
ofrobots: done.

PR #40, opened 2015-11-10
get the e2e test functional again

ofrobots:
- work around an regression introduced in acorn version 2.6.0
- switch to a promise based test
- remove stale files and dependencies
- Promises. How do they work?


ofrobots:
@matthewloring PTAL.

test/e2e/test.js
```diff
@@ -20,14 +20,13 @@ function fib(n) {
  * limitations under the License.
  */
 
-process.env.GCLOUD_INSIGHTS_LOGLEVEL=4;
-process.env.GCLOUD_DEBUG_ENABLED=true;
+process.env.GCLOUD_DEBUG_LOGLEVEL=3;
 
 var assert = require('assert');
-var util = require('util');
 var GoogleAuth = require('google-auth-library');
```
matthewloring: Is there a reason why we need the dependency on google-auth-library here? We have corresponding functionality in diagnostics common which should already be used by this package.

ofrobots: Historical baggage. It could be a cleanup task, but isn't essential for getting the test-case functional again.

ofrobots:
For posterity, here's the acorn issue: https://github.com/ternjs/acorn/issues/354

test/e2e/test.js
```diff
@@ -35,118 +34,169 @@ var SCOPES = [
...
+          return a.name === 'n';
+        });
+        assert.ok(arg, 'should find the n argument');
+        assert.strictEqual(arg.value, '10');
+        console.log('Test passed');
+        return true;
+      })
+      .catch(function(err) {
+        console.error(err);
+        process.exit(2);
```
matthewloring: Is there a reason why we catch these and exit with failure as opposed to letting them throw out of the promise?

matthewloring:
Promises :fearful: LGTM with question

test/e2e/test.js
ofrobots: I can let it throw out of the promise. I was originally contemplating doing something with the error, but this doesn't seem necessary anymore.

ofrobots: Done.

PR #41, opened 2015-11-10
run e2e test by default

ofrobots:
@matthewloring PTAL.

### DIFF
--- a/bin/run-e2e.sh
+++ b/bin/run-e2e.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd test/e2e
+
+if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots Can you add a check that $encrypted_a8f6bb4bf8ae_key and $encrypted_a8f6bb4bf8ae_iv are defined before decryption in the before_install step?

ofrobots:
@matthewloring PTAL.

bin/run-e2e.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd test/e2e
+
+if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then
```
matthewloring: Do we need to recheck the condition here?

matthewloring:
LGTM (we should stop running against 2)

PR #45, opened 2015-11-11
fix intermittently failing testcase

ofrobots:
@matthewloring PTAL.

### DIFF
--- a/test/test-debugletapi.js
+++ b/test/test-debugletapi.js
@@ -204,31 +204,6 @@ describe('Debuglet API', function() {
           done();
         });
     });
-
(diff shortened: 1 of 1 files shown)


test/test-debugletapi.js
```diff
@@ -206,7 +206,7 @@ describe('Debuglet API', function() {
     });
 
     it('should not tie down event loop on large breakpoints', function (done) {
```
matthewloring: This test appears to test the time it takes to stringify the object. If it doesn't test our application logic can we get rid of it?

matthewloring:
LGTM

Issue #37, opened 2015-11-05
Support transpiled JavaScript


ofrobots:
We already support source-maps for coffeescript, but we haven't tested with transpiled JavaScript.

matthewloring:
Fixed by https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/159.

PR #48, opened 2015-11-19
Added source map support for babel

matthewloring:
For now, we disallow babel pre-compiled files with the js extension
because we have no other way to differentiate pre and post compiled
files. Supported extensions are es6, es, jsx.

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -331,6 +331,14 @@ module.exports.create = function(logger_, config_, fileStats_) {
               uncompiled + ' <<', err);
           }
         };
+      case 'es6':
+      case 'es':
+      case 'jsx':
+        return function(uncompiled) {
+          // If we want to support es6 watch expressions we can compile them here.
+          // Babel is a large dependency to have if we don't need it in all cases.
+          return uncompiled;
+        };
     }
     return null;
   }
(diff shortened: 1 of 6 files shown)


matthewloring:
@ofrobots PTAL.

lib/v8debugapi.js
```diff
@@ -331,6 +331,14 @@ module.exports.create = function(logger_, config_, fileStats_) {
               uncompiled + ' <<', err);
           }
         };
+      case 'es6':
```
ofrobots: How likely is it that the users are naming their es6 source files with these extensions instead of just `.js`? If I were to do a project with Babel, I would like my source files to have the `.js` suffix. But then again, I haven't done anything with Babel yet, so I don't know what the standard practice is.

matthewloring: These extensions were pulled from the babel documentation. I don't know in practice which extensions people. We could implement setting breakpoints in compiled files where both source and target extensions are js but it may be worth waiting for that to be requested before adding the complication here.

ofrobots:
I'm fine with landing this, but the title needs to be updated as this doesn't solve the full use-case of supporting transpiled JS.

ofrobots:
LGTM w/ comment.

matthewloring:
Progress on #37

PR #49, opened 2015-11-30
fix reporting of file not found error

ofrobots:
The callback was getting called with the wrong value resulting the breakpoint
never being updated w/ the error on the server side. This causes the client to
wait upon a breakpoint forever even though we didn't even set the breakpoint in
the debuglet.


ofrobots:
@matthewloring ptal.

matthewloring:
Is there a way that the e2e tests could be expanded to make sure this path is tested somehow?

matthewloring:
Just saw #50, LGTM

Issue #51, opened 2015-11-30
Node.js apps that are sub-directories of the git repository don't work.


ofrobots:
Scenario:

1. Git repository with a directory Server/ that contains a Node.js app. i.e. `Server/app.js` and `Server/app.yaml` exist.
1. Host the repository in a gcloud source repo.
1. Deploy the app, or copy `Server/*` to `/tmp/app` and run it locally from there.

Once deployed the path to app.js is `/app/app.js`. But to the Debug UI, this is `Server/app.js`.

Immediate goal: in this scenario we should be able to set a breakpoint in app.js without being confused.

Secondary goal: consider the following directory layout:
`Server/app.js` -> `/app/app.js`
`Server/foo/app.js`-> `/app/foo/app.js`
`Server/app/app.js` -> `/app/app/app.js`

In such an app, we should be able to set the breakpoint in the correct app.js when the server sets a breakpoint in `Server/app.js`, `Server/app/app.js`. We should report an error when the server asks for a breakpoint in `app.js`. We need test cases for these scenario.

matthewloring:
The concern is that they have a directory structure such as:

`Server/app.js` -> `/app.js`
`Server/Server/app.js` -> `/Server/app.js`

When requesting to set a breakpoint in `Server/app.js` we will incorrectly attribute it to the code that was copied from `Server/Server/app.js`. If we could see all of the code on the server we could construct a map from server file names to executing file names but I don't believe it is possible to construct this map soundly in all cases starting only with a single server file name. It would be possible to build the map if we assume that path components are not repeated (such as nested directories with the name `Server`) but we would be unable to detect when this assumption is violated and would silently break and do the wrong thing. Alternatively, we could ask the user to provide us the mapping that should be used (for the above example they could provide `Server/` -> `/`) but this would help from the UI as well as a syntax for expressing mappings.

For now, I can add a heuristic that checks to see if there is only a single file with the provided name regardless of path but this will only be a slight improvement over the existing logic. Let me know what you think about the above options.

ofrobots:
>  It would be possible to build the map if we assume that path components are not repeated (such as nested directories with the name Server)

I don't think we can assume this.

[...]

PR #52, opened 2015-11-30
Heuristic for resolving ambiguous file paths

matthewloring:
This heuristic will correctly find scripts if their name is unique in
the executing project regardless of path to handle the case where the
app was deployed to a different path than that used in development.

Progress on #51

### DIFF
--- a/config.js
+++ b/config.js
@@ -20,6 +20,11 @@ module.exports = {
   enabled: true,
   workingDirectory: process.cwd(),
 
+  // The path within your repository to the directory containing the
+  // package.json for your deployed application. This should be provided
+  // if your deployed application appears as a subdirectory of your repository.
+  appPathRelativeToRepository: undefined,
+
   // Log levels: 0-disabled,1-error,2-warn,3-info,4-debug.
   logLevel: 1,
 
--- a/index.js
+++ b/index.js
@@ -31,6 +31,10 @@ if (process.env.hasOwnProperty('GCLOUD_DEBUG_LOGLEVEL')) {
 if (process.env.hasOwnProperty('GCLOUD_DEBUG_DISABLE')) {
   config.enabled = false;
 }
+if (process.env.hasOwnProperty('GCLOUD_DEBUG_REPO_APP_PATH')) {
+  config.appPathRelativeToRepository =
+    process.env.GCLOUD_DEBUG_REPO_APP_PATH;
+}
 
 if (config.enabled) {
   var debuglet = new Debuglet(
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -384,10 +384,26 @@ module.exports.create = function(logger_, config_, fileStats_) {
   // }
 
   function findScripts(scriptPath) {
+    // Use repository relative mapping if present.
+    if (config.appPathRelativeToRepository) {
+      var candidate = scriptPath.replace(config.appPathRelativeToRepository,
+        config.workingDirectory);
+      // There should be no ambiguity resolution if project root is provided.
+      return fileStats[candidate] ? [ candidate ] : [];
+    }
     var regexp = pathToRegExp(scriptPath);
-    var matches = Object.keys(fileStats).filter(function(script) {
-      return regexp.test(script);
-    });
+    // Next try to match path.
+    var matches = Object.keys(fileStats).filter(regexp.test.bind(regexp));
+    // Finally look for files with the same name regardless of path.
+    if (matches.length !== 1) {
+      matches = Object.keys(fileStats);
+      var components = scriptPath.split(path.sep);
+      for (var i = components.length - 1;
+           i >= 0 && matches.length > 1; i--) {
+        regexp = pathToRegExp(components.slice(i).join(path.sep));
+        matches = matches.filter(regexp.test.bind(regexp));
+      }
+    }
     return matches;
   }
 
--- a/test/test-v8debugapi.js
+++ b/test/test-v8debugapi.js
@@ -80,6 +80,28 @@ describe('v8debugapi', function() {
       });
     });
 
+  it('should disambiguate incorrect path if filename is unique',
+    function(done) {
+      require('./fixtures/foo.js');
+      var bp = { id: 0, location: {line: 1, path: '/test/foo.js'}};
+      api.set(bp, function(err) {
+        assert.ifError(err);
+        api.clear(bp);
+        done();
+      });
+    });
+
+  it('should disambiguate incorrect path if partial path is unique',
+    function(done) {
+      require('./fixtures/foo.js');
+      // hello.js is not unique but a/hello.js is.
+      var bp = { id: 0, location: {line: 1, path: '/Server/a/hello.js'}};
+      api.set(bp, function(err) {
+        assert.ifError(err);
+        api.clear(bp);
+        done();
+      });
+    });
 
   describe('invalid breakpoints', function() {
     var badBreakpoints = [
--- a/test/standalone/test-env-relative-repository-path.js
+++ b/test/standalone/test-env-relative-repository-path.js
@@ -1,0 +1,58 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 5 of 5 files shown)


matthewloring:
@ofrobots This implements the heuristic described in #51. PTAL.

matthewloring:
@ofrobots This PR has been updated to read git root directory from an environment variable and correct breakpoint file paths appropriately. PTAL.

config.js
```diff
@@ -20,6 +20,11 @@ module.exports = {
   enabled: true,
   workingDirectory: process.cwd(),
 
+  // The root directory containing the package.json for your project
+  // inside your git repository. This should be provided if your deployed
+  // application appears as a subdirectory of your git repository.
+  gitRoot: undefined,
```
ofrobots: The explanation of what this is wasn't obvious upon reading the comment. Is this the absolute path of my git directory? Is it the absolute path of the directory containing the `package.json`? Don't you mean to ask for the relative path of the package.json in the source-repository. Also, we don't have a dependency on the source repository being git specifically – there is no reason, in principle why it has to be git.

Also, we probably need a longer explanation on why this value should be provided. Whether we do it in this comment could be debatable.

A better name for this variable would be something like 'appPathRelativeToRepository.

matthewloring: I've taken another stab at the wording here. I think we'll need to document this more completely in the README once we're satisfied with the behavior here.

matthewloring:

index.js
```diff
@@ -31,6 +31,10 @@ if (process.env.hasOwnProperty('GCLOUD_DEBUG_LOGLEVEL')) {
 if (process.env.hasOwnProperty('GCLOUD_DEBUG_DISABLE')) {
   config.enabled = false;
 }
+if (process.env.hasOwnProperty('GCLOUD_DEBUG_RELATIVE_REPOSITORY_PATH')) {
```
ofrobots: Is this config important enough to warrant an config and an environment variable? I don't expect people will change this from run-to-run, so a config option is probably enough.

matthewloring: I'm happy to get rid of the environment variable, the hope was to let people configure it with whatever mechanism is most convenient since it doesn't introduce undue complexity to support both. If you'd like it to be config only I can remove the env var.

ofrobots: I forget that debug doesn't currently have a mechanism to accept config on startup. We'll have to keep the env. var. until we can fix that.

lib/v8debugapi.js
```diff
@@ -384,10 +384,27 @@ module.exports.create = function(logger_, config_, fileStats_) {
   // }
 
   function findScripts(scriptPath) {
+    // Use repository relative mapping if present
+    if (config.appPathRelativeToRepository) {
+      var candidate = scriptPath.replace(config.appPathRelativeToRepository,
+        config.workingDirectory);
```
ofrobots: Did you conscious choose to replace with `config.workingDirectory`? The alternative would be to replace with an empty string. The former assumes that application root == working directory. Thoughts?

matthewloring: This was a conscious decision, I was hoping to support the use case you originally discussed in the bug:

```javascript
Server/app.js -> /app/app.js
Server/foo/app.js-> /app/foo/app.js
Server/app/app.js -> /app/app/app.js
```
If they happen to deploy their server directory to /app we want to support that too.

lib/v8debugapi.js
```diff
@@ -384,10 +384,27 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+    }
+    // Next try to match path
     var regexp = pathToRegExp(scriptPath);
     var matches = Object.keys(fileStats).filter(function(script) {
       return regexp.test(script);
     });
+    // Finally look for files with the same name regardless of path
+    if (matches.length === 0) {
+      var splitQuery = scriptPath.split(path.sep);
+      scriptPath = splitQuery[splitQuery.length - 1];
```
ofrobots: `path.basename` might be useful here.

matthewloring: That makes things a lot cleaner!

lib/v8debugapi.js
```diff
@@ -384,10 +384,25 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+      return fileStats[candidate] ? [ candidate ] : [];
+    }
+    // Next try to match path
     var regexp = pathToRegExp(scriptPath);
     var matches = Object.keys(fileStats).filter(function(script) {
       return regexp.test(script);
     });
+    // Finally look for files with the same name regardless of path
+    if (matches.length === 0) {
+      scriptPath = path.basename(scriptPath);
```
ofrobots: Another option would be to peel one directory at a time to see if we can get a unique match. For example, I may have `Server/app/index.js` and `Server/foo/index.js`. Just looking for `index.js` will find duplicates, even though we ought to be able to do better here.

index.js
ofrobots: I don't like env. var name too much, but I don't have obviously better suggestions at this point.

matthewloring: How about `GCLOUD_DEBUG_REPO_APP_PATH`?

ofrobots: Sgtm

matthewloring: Done.

lib/v8debugapi.js
matthewloring: We now do a reverse search through the path components refining a list of candidates each iteration to find the right file in this situation.

lib/v8debugapi.js
```diff
@@ -385,9 +385,29 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
-    var matches = Object.keys(fileStats).filter(function(script) {
+    var filterScripts = function(script) {
       return regexp.test(script);
-    });
+    };
+    // Use repository relative mapping if present
+    if (config.appPathRelativeToRepository) {
+      var candidate = scriptPath.replace(config.appPathRelativeToRepository,
+        config.workingDirectory);
+      // There should be no ambiguity resolution if project root is provided
```
ofrobots: style-guide nit: add '.' at the end of the comment (here an elsewhere on the newly added comments)

lib/v8debugapi.js
```diff
@@ -385,9 +385,29 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+      return fileStats[candidate] ? [ candidate ] : [];
+    }
+    // Next try to match path
+    var matches = Object.keys(fileStats).filter(filterScripts);
+    // Finally look for files with the same name regardless of path
+    if (matches.length !== 1) {
+      matches = Object.keys(fileStats);
+      var components = scriptPath.split(path.sep);
+      var i = components.length - 1;
```
ofrobots: A for loop would be better here, stylistically.

lib/v8debugapi.js
```diff
@@ -385,9 +385,28 @@ module.exports.create = function(logger_, config_, fileStats_) {
 
   function findScripts(scriptPath) {
     var regexp = pathToRegExp(scriptPath);
-    var matches = Object.keys(fileStats).filter(function(script) {
+    var filterScripts = function(script) {
       return regexp.test(script);
```
ofrobots: The closure-captured `regexp` is mutable. It would be better to pass it as an argument, or possibly, just remove filterScripts completely just bind and pass `regexp.test` to `Array.prototype.filter`.

lib/v8debugapi.js
```diff
@@ -385,9 +385,28 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+      return fileStats[candidate] ? [ candidate ] : [];
+    }
+    // Next try to match path.
+    var matches = Object.keys(fileStats).filter(filterScripts);
+    // Finally look for files with the same name regardless of path.
+    if (matches.length !== 1) {
+      matches = Object.keys(fileStats);
+      var components = scriptPath.split(path.sep);
+      for (var i = components.length - 1;
+          i >= 0 && matches.length > 1; i--) {
```
ofrobots: align `i` with `var`

lib/v8debugapi.js
matthewloring: Binding `regexp.test` directly is definitely cleaner!

lib/v8debugapi.js

test/test-v8debugapi.js
```diff
@@ -80,6 +80,16 @@ describe('v8debugapi', function() {
...
+  it('should disambiguate incorrect path if filename is unique',
+    function(done) {
+      require('./fixtures/foo.js');
+      var bp = { id: 0, location: {line: 1, path: '/test/foo.js'}};
+      api.set(bp, function(err) {
+        assert.ifError(err);
+        api.clear(bp);
+        done();
+      });
+    });
```
ofrobots: Add another test that exercises the for loop in `findScripts`. Basically, have multiple `foo.js` files in different directories.

ofrobots:
LGTM w/ comments.

test/test-v8debugapi.js
matthewloring: Test added using `hello.js` which exists in the fixtures root as well as the `a` and `b` subdirectories. We can successfully set a breakpoint in `/Server/a/hello.js` despite `/Server/` being incorrect and not part of the actual path.

PR #54, opened 2015-12-04
Setting a bp in json should give correct error

### LINKED ISSUE #53: Setting breakpoint in package.js gives sourcemap not found error

matthewloring:
Fixes #53

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -28,8 +28,10 @@
 
 /** @const */ var messages = {
   INVALID_BREAKPOINT: 'invalid breakpoint - id or location missing',
-  BREAKPOINT_ONLY_SUPPORTS_JAVASCRIPT: 'Only files with .js extensions are supported',
-  SOURCE_FILE_NOT_FOUND: 'A script matching the source file was not found loaded on the debuggee',
+  BREAKPOINT_ONLY_SUPPORTS_JAVASCRIPT:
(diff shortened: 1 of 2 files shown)


matthewloring:
@ofrobots PTAL.

lib/v8debugapi.js
```diff
@@ -149,7 +150,8 @@ module.exports.create = function(logger_, config_, fileStats_) {
                 breakpoint.location.line = mappedPos.line;
               } catch (err) {
                 return setErrorStatusAndCallback(cb, breakpoint,
-                  StatusMessage.CONDITION, messages.SOURCE_MAP_READ_ERROR + err);
+                  StatusMessage.BREAKPOINT_CONDITION,
```
ofrobots: Ooh, this was a typo before?

ofrobots: I think `BREAKPOINT_SOURCE_LOCATION` might be a better choice here. WDYT?

matthewloring: Unfortunately it was :frown: `BREAKPOINT_SOURCE_LOCATION` sounds good to me!

ofrobots: In that case, it sounds like we are missing a test-case. Can you add a test-case for this and other error returns from this function?  separate PR is fine.

ofrobots:
LGTM.

PR #55, opened 2015-12-04
fix the script path we use to set the v8 bp

ofrobots:
We were using the original path provided by the Debug API rather than the app
local path we computed based on the file-matching heuristics.

@matthewloring PTAL.

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -286,7 +286,7 @@ module.exports.create = function(logger_, config_, fileStats_) {
       column += MODULE_WRAP_PREFIX_LENGTH - 1;
     }
 
-    var v8bp = setByRegExp(scriptPath, line, column);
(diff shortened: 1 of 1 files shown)


matthewloring:
:fearful: Scary that this made it so far. Maybe the e2e tests could be augmented to cover this case. LGTM.

ofrobots:
It does have to be a test-case that moves files in order to make the bug to manifest. Either an e2e test, or at least, a slightly more advanced test case is needed here. Opened #56.

matthewloring:
@ofrobots Does the [relative repository paths test](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/test/standalone/test-env-relative-repository-path.js) cover this case? It sets a breakpoint at a non-existent path that is normalized using the repo root hint from the environment variable so if the incorrect path were being used this test should fail.

PR #58, opened 2015-12-07
Support breakpoints inside node_modules

### LINKED ISSUE #57: Support breakpoints for files inside node_modules

matthewloring:
Now only the .git directory is ignored when scanning the local
filesystem.

Fixes #57

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -26,7 +26,7 @@ function run {
     C="$(npm bin)/istanbul cover --dir ./coverage/${counter}"
     ((counter++))
   fi
-  ($C "$(npm bin)/_mocha" -- $* --timeout 4000 --R spec) || exit 1
+  ($C "$(npm bin)/_mocha" -- $* --timeout 15000 --R spec) || exit 1
 }
 
 # Run test/coverage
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
Hmm.. The issue with this approach is that two instances of the app deployed at slightly different times (or re-deployed, restarted etc.) might end up with different npm modules getting installed. Since we will take the node_modules into account when computing the unique-id of the application source, these would look like different debug targets, with the user having no idea what's different between them.

matthewloring:
This is a concern. If the user intends to set breakpoints inside node_modules then these should be treated as separate debug targets. We could add a config option along the lines of 'optimize for no breakpoints in node_modules' but the issue is that you often don't know where you want to set a breakpoint until things start falling apart and we wouldn't want to strand users in a state where they are unable to set breakpoints where they need to because their deployed node_modules has a different file structure. This is definitely not a good thing in the common case where people aren't setting breakpoints in node_modules though. Do you have any thoughts?

matthewloring:
closing for now.

tcf909:
Any chance of reopening this? 

I think debugging dependencies is of great value. We ourselves package much of our own code has separate modules to keep things as DRY as possible. Setting up a seperate agent for each module is quite the pain.

Couldn't we use the "includeNodeModules" option to allow the agent to instrument the dependencies? That way its off by default but allows some to use it when appropriate?

PR #62, opened 2015-12-19
Don't repeat frame arguments in frame locals

### LINKED ISSUE #59: When a breakpoint is hit inside an express handler, two variables named res are captured.

matthewloring:
It appears that function arguments captured in closures show up as both
an argument and a local to v8. This patch stops us from double reporting
in such cases.

Fixes #59


matthewloring:
@ofrobots PTAL.

matthewloring:
@ofrobots PTAL.

ofrobots:
lgtm.

PR #63, opened 2015-12-28
Update dev dependencies

matthewloring:
I was playing around with snyk and found
https://snyk.io/vuln/npm:uglify-js:20151024. We depend on uglify
through istanbul (dev only) so it shouldn't matter but
updating is always good as are green badges.


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #65, opened 2016-01-06
Unify warnings for using node v5.2 and <v0.12 through debugger ui

matthewloring:
The agent is currently crashing on node 0.10 because
EventEmitter.prototype is undefined which breaks inheritance for the
debuglet. This patch causes the agent to shutdown immediately if running
on an unsupported version of node. It also addresses #7 since the
version check runs first and it is now easier to reason about graceful
exit in this case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,6 @@
 var config = require('./config.js');
 var logger = require('@google/cloud-diagnostics-common').logger;
 var Debuglet = require('./lib/debuglet.js');
-var semver = require('semver');
 
 // exports is populated by the agent
 module.exports = {};
@@ -38,11 +37,6 @@ if (process.env.hasOwnProperty('GCLOUD_DEBUG_REPO_APP_PATH')) {
 }
 
 var log = logger.create(config.logLevel, '@google/cloud-debug');
-
-if (semver.satisfies(process.version, '5.2')) {
-  log.error('Debug is not supported on Node v5.2');
-  return;
-}
 
 if (config.enabled) {
   var debuglet = new Debuglet(config, log);
--- a/lib/debuglet.js
+++ b/lib/debuglet.js
@@ -18,8 +18,9 @@
 
 var fs = require('fs');
 var path = require('path');
-var EventEmitter = require('events');
+var EventEmitter = require('events').EventEmitter;
 var util = require('util');
+var semver = require('semver');
 
 var v8debugapi = require('./v8debugapi.js');
 var DebugletApi = require('./debugletapi.js');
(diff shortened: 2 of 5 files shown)


matthewloring:
The other option I see here is to add special logic in the debuglet to handle the case where it is running in 0.10. I would rather shut down the agent immediately in this case as it communicates the issue to the user more immediately than waiting until a breakpoint set is attempted. @ofrobots WDYT?

index.js
```diff
@@ -39,11 +38,13 @@ if (process.env.hasOwnProperty('GCLOUD_DEBUG_REPO_APP_PATH')) {
 
 var log = logger.create(config.logLevel, '@google/cloud-debug');
 
-if (semver.satisfies(process.version, '5.2')) {
-  log.error('Debug is not supported on Node v5.2');
-  return;
+if (semver.satisfies(process.version, '5.2 || <0.12')) {
+  log.error('Debug is not supported on Node v5.2 or <v0.12');
+  process.exit(1);
```
ofrobots: I am not so sure about the `process.exit`. It is quite possible for users to have dev environments different from the version of Node they would get on deployment. We are converting a lower severity problem ('my debug agent isn't starting') to a very high severity problem ('my app doesn't deploy').

ofrobots:
I would be fine with this change if only we would use `return` instead of `process.exit` on unsupported versions.

index.js
matthewloring: That makes sense. I've downgraded back to a `return`.

matthewloring:
This has been updated to report all version failures at the time the first breakpoint is set as per offline discussion. Additionally, we no longer attempt to test on 0.10 as the package is not installable without npm support for namespaces. @ofrobots PTAL.

lib/debuglet.js
```diff
@@ -30,9 +31,9 @@ var StatusMessage = require('./apiclasses.js').StatusMessage;
 
 var assert = require('assert');
 
-var NODE_VERSION_MESSAGE = 'V8 Debug API not available – make sure you are ' +
-    'running Node.js >=0.12. Cloud Debugger agent will not be able to take ' +
-    'snapshots.';
+var NODE_VERSION_MESSAGE = 'An unsupported version of Node.js is running. ' +
+  'Cloud Debugger is only supported with Node.js >=0.12 with the exception of ' +
+  ' 5.2.';
```
ofrobots: How about rewording this to 'Node.js version not supported. Node.js 5.2.0 and versions older than 0.12 are not supported.'

test/e2e/test-error-message.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
ofrobots: 2016

ofrobots:
test/.DS_store should not be committed to the repository. Does our .gitignore not prevent this from happening already?

Secondly, our README.md needs to mention that 5.2.0 is not suppoted.

LGTM w/ these comments.

test/e2e/test-error-message.js
matthewloring: Done.

lib/debuglet.js
matthewloring: Sounds good.

matthewloring:
The .gitignore has been updated and the other comments have been addressed.

PR #66, opened 2016-01-07
Clean up coffeescript error messages

### LINKED ISSUE #2: CoffeeScript error messages don't make sense
The following is the output from the 'should capture without values for invalid watch expressions in compiled code' test (if the log level is changed to INFO):

```
ERROR:: Unable to compile break or watch point >> :) << { [SyntaxError: unmatched )]
  location: { first_line: 0, first_column: 8, last_column: 8 },
  toString: [Function],
  code: '0 || (:))',
  filename: undefined }
ERROR:: Unable to compile break or watch point >> process=this << [TypeError: Cannot read property '1' of null]
ERROR:: Unable to compile break or watch point >> ((x) -> x x) n << [TypeError: Cannot read property '1' of null]
ERROR:: Unable to compile break or watch point >> return << { [SyntaxError: cannot use a pure statement in an expression]
  location: { first_line: 0, first_column: 6, last_line: 0, last_column: 11 },
  toString: [Function],
  code: '0 || (return)',
  filename: undefined }
      ✓ should capture without values for invalid watch expressions in compiled code (207ms)
```

Apart from the first error, the remainder aren't very useful, but then again, I don't know CoffeeScript. We should look into these to make sure they make sense to CoffeeScripters.

matthewloring:
If compilation of a coffeescript expression fails, we will report the
associated error if the coffeescript compiler provides one. Otherwise,
we will just report that the compilation failed.

Fixes #2

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -36,6 +36,7 @@
   V8_BREAKPOINT_ERROR: 'Unable to set breakpoint in v8',
   SYNTAX_ERROR_IN_CONDITION: 'Syntax error in condition: ',
   ERROR_EVALUATING_CONDITION: 'Error evaluating condition: ',
+  ERROR_COMPILING_CONDITION: 'Error compiling condition.',
   DISALLOWED_EXPRESSION: 'Expression not allowed',
   SOURCE_MAP_URL_NOT_FOUND: 'The source map url could not be found in the compiled file',
   SOURCE_MAP_READ_ERROR: 'The source map could not be read or was incorrectly formatted',
@@ -132,7 +133,15 @@ module.exports.create = function(logger_, config_, fileStats_) {
         };
         compile = getBreakpointCompiler(breakpoint);
         if (breakpoint.condition && compile) {
-          breakpoint.condition = compile(breakpoint.condition);
+          try {
+            breakpoint.condition = compile(breakpoint.condition);
+          } catch (e) {
+            logger.info('Unable to compile condition >> ' +
+              breakpoint.condition + ' <<');
+            return setErrorStatusAndCallback(cb, breakpoint,
+                StatusMessage.BREAKPOINT_CONDITION,
+                messages.ERROR_COMPILING_CONDITION);
+          }
         }
         // TODO: more robust file finding of compiled files
         scriptPath = scriptPath.substr(0, scriptPath.lastIndexOf('.')) + '.js';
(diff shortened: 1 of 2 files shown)


matthewloring:
Example log output for invalid conditions test case:

```
INFO :: Unable to compile break or watch point >> :) << { [SyntaxError: unmatched )]
  location: { first_line: 0, first_column: 8, last_column: 8 },
  toString: [Function],
  code: '0 || (:))',
  filename: undefined }
INFO :: Unable to compile break or watch point >> process=this <<
INFO :: Unable to compile break or watch point >> ((x) -> x x) n <<
INFO :: Unable to compile break or watch point >> return << { [SyntaxError: cannot use a pure statement in an expression]
  location: { first_line: 0, first_column: 6, last_line: 0, last_column: 11 },
  toString: [Function],
  code: '0 || (return)',
  filename: undefined }
```

matthewloring:
@ofrobots PTAL.

lib/v8debugapi.js
```diff
@@ -328,8 +328,13 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
             // Strip out coffeescript scoping wrapper to get translated condition
             var re = /\(function\(\) {\s*0 \|\| \((.*)\);\n\n\}\)\.call\(this\);/;
-            var match = re.exec(compiled)[1];
-            return match ? match.trim() : match;
+            var match = re.exec(compiled);
+            if (match && match.length > 1) {
+              return match[1].trim();
+            } else {
+              logger.info('Unable to compile break or watch point >> ' +
+                uncompiled + ' <<');
```
ofrobots: The else path lacks a return value.

ofrobots:
I think we should go further. We have detected that the expression doesn't compile. We should figure out a way to report it to the debuglet-controller API so that the user can be informed too. We do know that the user isn't necessarily looking at the logs.

lib/v8debugapi.js
matthewloring: Fixed.

matthewloring:
I've updated both the watch expression and breakpoint condition paths to report compilation failures to the UI.

ofrobots:
Looking good. Can we add a test that ensures that the API gets the error message upon compilation failure for both conditions and watch expressions?

matthewloring:
Added tests.

ofrobots:
lgtm.

PR #68, opened 2016-01-08
Regression test for #56 on GH

### LINKED ISSUE #56: Need a test-case to catch repo issues like in #55
See: #55.

matthewloring:
When testing relative repository paths, we need to ensure that the set
breakpoint is hit when the script executes since v8 will accept
breakpoints in scripts that are not yet loaded masking an error in
selecting the correct script.

Fixes #56

### DIFF
--- a/test/fixtures/a/hello.js
+++ b/test/fixtures/a/hello.js
@@ -1,2 +1,4 @@
 'use strict';
-module.exports.hello = function() { console.log('hello world from a'); };
+module.exports.hello = function() {
+  console.log('hello world from a');
+};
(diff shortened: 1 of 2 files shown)


matthewloring:
@ofrobots PTAL.

ofrobots:
So the test was only checking whether we were able to add a breakpoint to v8 rather than checking if we were actually hitting the breakpoint.

Why did you have to change the breakpoint from `a/hello.js`?

matthewloring:
I was just hoping to leave the fixture unchanged and I can't set a breakpoint inside the function `a/hello.js` exports since it is on the same line as the export definition. I've updated the test to modify the fixture and keep `a/hello.js`.

ofrobots:
LGTM.

PR #69, opened 2016-01-08
Test for source context

### LINKED ISSUE #15: Need tests for source-context.json

matthewloring:
This ensures that the source context read from the source-context.json
file in the working directory is passed to the api when the debuggee
registers.

Fixes #15.


matthewloring:
@ofrobots PTAL.

test/standalone/test-debuglet.js
```diff
@@ -95,6 +96,39 @@ describe(__filename, function(){
     debuglet.start();
   });
 
+  it('should pass source context to api if present', function(done) {
+    fs.writeFileSync('source-context.json', '{ "a": 5 }', 'utf8');
```
ofrobots: Would it be possible to make this test state-less? i.e. run it from a directory where a source-context.json already exists? Alternatively, does the config for debug accept a working directory – if so point it to the directory with the source-context.json.

The benefit of stateless tests is that spurious IO issues (stale files from previous runs, or failure to write out these files) don't interfere with the test.

test/standalone/test-debuglet.js
```diff
@@ -95,6 +96,39 @@ describe(__filename, function(){
...
+    debuglet.once('started', function() {
+      debuglet.debugletApi_.request_ = request; // Avoid authing.
+    });
+    debuglet.once('registered', function(id) {
+      assert(id === 'foo');
+      debuglet.stop();
+      scope.done();
+      fs.unlinkSync('source-context.json');
+      done();
+    });
```
ofrobots: This pattern of code is showing up with regularity. Do you think it is worth refactoring this?

test/standalone/test-debuglet.js
matthewloring: Switched over to use a permanent `source-context.json` in `fixtures`.

test/standalone/test-debuglet.js
matthewloring: Refactored a few things in these tests to reduce duplication between the cases.

matthewloring:
@ofrobots Ready for a look.

ofrobots:
lgtm

PR #70, opened 2016-01-11
Regression test for #50 on GH

### LINKED ISSUE #50: Need tests for the debuglet's interaction with the service
Tests would have caught e.g. #49 .

matthewloring:
Ensures that errors occuring when a breakpoint is set are reported to
the api.

Fixes #50


matthewloring:
@ofrobots PTAL.

ofrobots:
lgtm.

Issue #71, opened 2016-01-13
Investigate breakpoint capture performance 


matthewloring:

matthewloring:
On my local machine using Node v5.3.x, breakpoints were captured in ~5-7ms depending on the size of watch expressions computed. Breakpoints were reported to the API in another 2-4ms for a total of 7-11ms. Unfortunately, on versions of Node before 1.6, variable resolution takes roughly 10x longer due to https://github.com/iojs/io.js/issues/1190. Unfortunately, on travis these captures take as much as 250ms and so the test introduced to monitor performance in #73 won't be run with unit tests on each PR.

matthewloring:
Closing this now that we have some data. We can revisit at a later time.

PR #73, opened 2016-01-14
Assorted performance improvements

matthewloring:
- Speeds up variable resolution 10x on node >1.6.
- Remove the event emitter from v8debugapi in favor of directly invoking
  callbacks.
- Includes a test checking that large breakpoint capture doesn't go
  above 20ms on node 1.6+ and doesn't go above 110ms on node <1.6.

Addresses #71

### DIFF
--- a/lib/state.js
+++ b/lib/state.js
@@ -22,6 +22,7 @@ module.exports = {
 };
 
 var assert = require('assert');
+var semver = require('semver');
 
 var StatusMessage = require('./apiclasses.js').StatusMessage;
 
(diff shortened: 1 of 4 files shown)


matthewloring:
@ofrobots PTAL.

matthewloring:
@ofrobots This has been separated out into different commits per idea. Ready for another look.

ofrobots:
What would be needed to get this e2e test running on travis?

matthewloring:
Updated to use `that.config_.capture.maxProperties`. Unfortunately, I don't think the travis environment is performant enough to get reasonable numbers. We can set it up to run locally only though.

ofrobots:
LGTM w/ comment.

PR #74, opened 2016-01-14
Only mark breakpoints done when api call succeeds

matthewloring:
The first part of the TODO is deemed no longer necessary because we
already have our request handle retriest for transient errors.

### DIFF
--- a/lib/debuglet.js
+++ b/lib/debuglet.js
@@ -339,19 +339,16 @@ Debuglet.prototype.completeBreakpoint_ = function(breakpoint) {
   var that = this;
 
   that.logger_.info('\tupdating breakpoint data on server', breakpoint.id);
-  // TODO: in case of transient errors, retry the update operation
-  // Put in the completed breakpoints map only when the update successfully
-  // completes. Otherwise - we refuse to set the breakpoint, but the server
-  // has never seen the update.
   that.debugletApi_.updateBreakpoint(breakpoint, function(err/*, body*/) {
     if (err) {
       that.logger_.error('Unable to complete breakpoint on server', err);
+    } else {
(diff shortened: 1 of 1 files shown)


matthewloring:
@ofrobots PTAL.

lib/debuglet.js
```diff
@@ -339,19 +339,16 @@ Debuglet.prototype.completeBreakpoint_ = function(breakpoint) {
...
     if (err) {
       that.logger_.error('Unable to complete breakpoint on server', err);
+    } else {
+      // TODO(ofrobots): breakpoint may be holding on to a lot of data. We only
+      // need to remember the breakpoint id.
+      that.completedBreakpointMap_[breakpoint.id] = breakpoint;
+      that.removeBreakpoint_(breakpoint);
     }
   });
-  // TODO(ofrobots): breakpoint may be holding on to a lot of data. We only
```
ofrobots: Here's the scenario:
- Server gets a breakpoint A
- We get notified about breakpoint A. We set the breakpoint. We hit the breakpoint.
- We update the breakpoint on the server. A network error happens so the server doesn't acknowledge that we have hit the breakpoint. We log an error, but continue.
- We get the list of breakpoints from the server. We find a breakpoint A that we have neither in the active nor completed lists.
- We set breakpoint A, we hit breakpoint A.

matthewloring: My understanding is that A will still be in the active list in step 4 above because it isn't removed until we hit `removeBreakpoint_` which only occurs after we receive a successful acknowledgement from the server.

ofrobots: SGTM.

ofrobots:
LGTM.

PR #78, opened 2016-01-14
Remove TODO to use v8 as parser

matthewloring:
It seems that this has been asked for in the past and the V8 team
justified why they do not want these apis to be exposed. See Andreas
Rossberg's answer here:
https://groups.google.com/forum/#!topic/v8-users/_WracRX9BTQ


matthewloring:
@ofrobots PTAL.

PR #79, opened 2016-01-15
Display error for native properties

matthewloring:
This prevents native properties from being reported as undefined when we
are unable to check their values for fear of executing stateful code.

### DIFF
--- a/lib/state.js
+++ b/lib/state.js
@@ -26,6 +26,10 @@ var semver = require('semver');
 
 var StatusMessage = require('./apiclasses.js').StatusMessage;
 
+var BUFFER_FULL_MESSAGE_INDEX = 0;
+var NATIVE_PROPERTY_MESSAGE_INDEX = 1;
+var GETTER_MESSAGE_INDEX = 2;
+
 // TODO: document this file
 
 // returns an object with three fields: stacksframes,
@@ -79,13 +83,17 @@ function StateResolver(execState, expressions, config) {
   this.evaluatedExpressions_ = [];
   this.totalSize_ = 0;
 
-  // The 0-th object in the variable table is a sentinel 'buffer full' error
-  // message value
-  this.rawVariableTable_ = [ null ];
-  this.resolvedVariableTable_ = [
+  this.rawVariableTable_ = [ null, null, null ];
+  this.resolvedVariableTable_ = [];
+  this.resolvedVariableTable_[BUFFER_FULL_MESSAGE_INDEX] =
     { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
-                                'Buffer full', true) }
-  ];
+                                'Max data size reached', true) };
+  this.resolvedVariableTable_[NATIVE_PROPERTY_MESSAGE_INDEX] =
+    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
+                                'Native properties are not available', true) };
+  this.resolvedVariableTable_[GETTER_MESSAGE_INDEX] =
+    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
+                                'Properties with getters are not available', true) };
 }
 
 
@@ -114,7 +122,7 @@ StateResolver.prototype.capture = function() {
   }
 
   // Now resolve the variables
-  var index = 1; // skip the sentinel value
+  var index = 3; // skip the sentinel values
   while (index < that.rawVariableTable_.length && // NOTE: length changes in loop
          that.totalSize_ < that.config_.capture.maxDataSize) {
     assert(!that.resolvedVariableTable_[index]); // shouldn't have it resolved yet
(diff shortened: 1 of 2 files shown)


matthewloring:
@ofrobots PTAL.

lib/state.js
```diff
@@ -95,6 +95,10 @@ StateResolver.prototype.capture = function() {
   // Gather the stack frames first
   var that = this;
   var frames = that.resolveFrames_();
+  this.nativeErrorMirror =
+    this.state_.frame(0).evaluate('\'Cannot display native properties.\'');
+  this.getterErrorMirror =
+    this.state_.frame(0).evaluate('\'Cannot display properties with getters.\'');
```
ofrobots: Why is the evaluate needed?

lib/state.js
```diff
@@ -95,6 +95,10 @@ StateResolver.prototype.capture = function() {
   // Gather the stack frames first
   var that = this;
   var frames = that.resolveFrames_();
+  this.nativeErrorMirror =
```
ofrobots: Since this is supposed to be a private property, add a `_` suffix.

lib/state.js
ofrobots: It would be preferable to return a `StatusMessage` indicating error rather than a string. See how the buffer full error message works:

```js
  // The 0-th object in the variable table is a sentinel 'buffer full' error
  // message value
  this.rawVariableTable_ = [ null ];
  this.resolvedVariableTable_ = [
    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
                                'Buffer full', true) }
```

matthewloring: I wasn't aware of this channel for communicating errors. This is much better than a string value.

lib/state.js
matthewloring: No longer relevant.

matthewloring:

lib/state.js
```diff
@@ -81,13 +85,17 @@ function StateResolver(execState, expressions, config) {
...
-  this.rawVariableTable_ = [ null ];
-  this.resolvedVariableTable_ = [
+  this.rawVariableTable_ = [ null, null, null ];
+  this.resolvedVariableTable_ = [];
+  this.resolvedVariableTable_[BUFFER_FULL_MESSAGE_INDEX] =
+    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
+                                'Buffer full', true) };
+  this.resolvedVariableTable_[NATIVE_PROPERTY_MESSAGE_INDEX] =
+    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
+                                'Native property', true) };
```
ofrobots: I wonder if we should make this more descriptive. Something like 'Native properties are not available'?

lib/state.js
```diff
@@ -81,13 +85,17 @@ function StateResolver(execState, expressions, config) {
...
+  this.resolvedVariableTable_[GETTER_MESSAGE_INDEX] =
     { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
-                                'Buffer full', true) }
```
ofrobots: likewise.

test/test-v8debugapi.js
```diff
@@ -332,7 +336,39 @@ describe('v8debugapi', function() {
...
+      api.set(bp, function(err) {
+        assert.ifError(err);
+        api.wait(bp, function(err) {
+          assert.ifError(err);
+
+          var procEnv = bp.evaluatedExpressions[0];
+          assert.equal(procEnv.name, 'process.env');
+          var envVal = bp.variableTable[procEnv.varTableIndex];
+          envVal.members.forEach(function(member) {
+            assert.equal(member.varTableIndex, 1);
```
ofrobots: It would be preferable to check if this is a StatusMessage with isError indicated rather than using a hard-coded index.

test/test-v8debugapi.js
```diff
@@ -332,7 +336,39 @@ describe('v8debugapi', function() {
...
+          });
+          var hasGetter = bp.evaluatedExpressions[1];
+          var getterVal = bp.variableTable[hasGetter.varTableIndex];
+          assert(getterVal.members.some(function(m) {
+            return m.value === '5';
+          }));
+          assert(getterVal.members.some(function(m) {
+            return m.varTableIndex === 2;
```
ofrobots: Again, check for error rather than a hard-coded index.

ofrobots:
LGTM w/ comments.

lib/state.js
matthewloring: Done.

lib/state.js
matthewloring: Done.

test/test-v8debugapi.js
matthewloring: Done.

test/test-v8debugapi.js
matthewloring: Done.

PR #80, opened 2016-01-15
Only compute hash if no other uid available

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/scanner.js
```diff
@@ -39,11 +39,12 @@ function scan(baseDir, callback) {
  * This function accept an array of filenames and computes a unique hash-code
  * based on the contents.
  *
+ * @param {Boolean} shouldHash whether a hash should be computed
  * @param {!Array<string>} fileList array of filenames
  * @param {!function(?Error, string, Object)} callback error-back style callback
  *    returning the hash-code and an object containing file statistics.
  */
-function computeStats(fileList, callback) {
+function computeStats(shouldHash, fileList, callback) {
```
ofrobots: It would be more idiomatic to use the following argument order: fileList, shouldHash, callback.

Also, since the argument order for the callback should be (err, fileStats, hash) with the optional value (hash as the last value).

lib/scanner.js
```diff
@@ -39,11 +39,12 @@ function scan(baseDir, callback) {
  * This function accept an array of filenames and computes a unique hash-code
  * based on the contents.
  *
+ * @param {Boolean} shouldHash whether a hash should be computed
  * @param {!Array<string>} fileList array of filenames
  * @param {!function(?Error, string, Object)} callback error-back style callback
```
ofrobots: `s/string/?string/`

lib/scanner.js
matthewloring: Fixed.

lib/scanner.js
matthewloring: Fixed.

matthewloring:

ofrobots:
LGTM if the tests pass.

PR #83, opened 2016-01-20
Store breakpoint id instead of breakpoint where possible

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM.

PR #85, opened 2016-01-21
Enable maxExpandFrames config option

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/state.js
```diff
@@ -177,14 +177,15 @@ StateResolver.prototype.resolveFrames_ = function() {
...
+      var resolveVars = i < this.config_.capture.maxExpandFrames;
+      frames.push(this.resolveFrame_(frame, resolveVars));
     }
   }
   return frames;
 };
 
 
-StateResolver.prototype.shouldFrameBeResolved_ = function(frame) {
+StateResolver.prototype.shouldFrameBeResolved_ = function(frame, frameNum) {
```
ofrobots: is the `frameNum` argument used?

matthewloring: Nope, removed.

lib/state.js
```diff
@@ -242,13 +243,18 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
 };
 
 
-StateResolver.prototype.resolveFrame_ = function(frame) {
-  var args = this.resolveArgumentList_(frame);
+StateResolver.prototype.resolveFrame_ = function(frame, resolveVars) {
+  var args = [];
```
ofrobots: It might be more succinct to write var args = resolveVars ? this.resolveArgumentList_(frame) : [];

test/test-v8debugapi.js
```diff
@@ -307,6 +307,31 @@ describe('v8debugapi', function() {
...
+        assert.ifError(err);
+        api.wait(bp, function(err) {
+          assert.ifError(err);
+          assert.ok(bp.stackFrames);
+          assert.ok(bp.variableTable);
+
+          var topFrame = bp.stackFrames[0];
+          assert.ok(topFrame);
+          assert.equal(topFrame['function'], 'foo');
+          assert.equal(topFrame.arguments.length, 0);
```
ofrobots: Instead of an empty list, I wonder if we can provide an error StatusMessage indicating that the argument and local list has been omitted. How hard would this be?

ofrobots:
LGTM once tests pass. Suggestion about error `StatusMessage` can be addressed in a follow on PR.

lib/state.js
matthewloring: Fixed.

test/test-v8debugapi.js
matthewloring: I've added the StatusMessage, @ofrobots please take another look.

lib/state.js
```diff
@@ -83,17 +100,11 @@ function StateResolver(execState, expressions, config) {
...
-                                'Max data size reached', true) };
-  this.resolvedVariableTable_[NATIVE_PROPERTY_MESSAGE_INDEX] =
-    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
-                                'Native properties are not available', true) };
-  this.resolvedVariableTable_[GETTER_MESSAGE_INDEX] =
-    { status: new StatusMessage(StatusMessage.VARIABLE_VALUE,
-                                'Properties with getters are not available', true) };
+  for (var i = 0; i < MESSAGE_TABLE.length; i++) {
+    this.resolvedVariableTable_[i] = MESSAGE_TABLE[i];
+  }
```
ofrobots: Would `Object.assign` work instead of doing a loop here?

lib/state.js
```diff
@@ -242,13 +254,27 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
...
+  } else {
+    args = [{
+      name: 'arguments_not_available',
+      varTableIndex: ARG_LOCAL_LIMIT_MESSAGE_INDEX
+    }];
+    locals = [{
+      name: 'locals_not_available',
+      varTableIndex: ARG_LOCAL_LIMIT_MESSAGE_INDEX
+    }];
+  }
```
ofrobots: If it doesn't get too unwieldy, I would still prefer a `args = resolveVars ? ... : ...` style initialization.

ofrobots:
LGTM w/ comments.

lib/state.js
matthewloring: Much cleaner, thanks.

lib/state.js
matthewloring: Yup, done.

ofrobots:
Ugh. You'll have to use `util._extend`. I forgot that `Object.assign` has only been there since V8 4.5.

PR #97, opened 2016-04-20
defer breakpoint callback

ofrobots:
Make sure the breakpoint hit callback isn't called back from the debug
listener, which swallows user exceptions thrown from the callback.

@matthewloring PTAL.

### DIFF
--- a/test/test-v8debugapi.js
+++ b/test/test-v8debugapi.js
@@ -700,6 +700,33 @@ describe('v8debugapi', function() {
         process.nextTick(function() {foo();});
       });
     });
+
+    it('should not silence errors thrown in the wait callback', function(done) {
+      var message = 'This exception should not be silenced';
+      // Remove the mocha listener.
+      var listeners = process.listeners('uncaughtException');
+      assert.equal(listeners.length, 1);
+      var originalListener = listeners[0];
+      process.removeListener('uncaughtException', originalListener);
+      process.once('uncaughtException', function(err) {
+        assert.ok(err);
+        assert.equal(err.message, message);
+        // Restore the mocha listener.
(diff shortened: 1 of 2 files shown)


matthewloring:
Could we add a test for this?

ofrobots:
Added test. PTAL.

test/test-v8debugapi.js
```diff
@@ -700,6 +700,31 @@ describe('v8debugapi', function() {
         process.nextTick(function() {foo();});
       });
     });
+
+    it('should not silence errors thrown in the wait callback', function(done) {
+      var message = 'This exception should not be silenced';
+      // Remove the mocha listener.
+      var originalListener = process.listeners('uncaughtException').pop();
```
matthewloring: Won't `pop` and `removeListener` attempt the same action? Also we may want to assert that there is only a single listener at this point to ensure we're always grabbing the mocha listener or that there aren't more attached by mocha we need to be aware of.

ofrobots: process.listeners returns a copy. Can't find the source quicky, but see here:
```
> var A = process.listeners('unhandledException')
undefined
> var B = process.listeners('unhandledException')
undefined
> A
[ [Function] ]
> B
[ [Function] ]
> A.pop()
[Function]
> A
[]
> B
[ [Function] ]
```

If there are multiple listeners, and the remaining ones don't abort, the test still works. If the remaining listeners abort, then that's fine – the test doesn't pass. But I agree that it would be simpler to add a test for single listener.

matthewloring: Ahh, those semantics are somewhat unexpected but that makes sense now.

matthewloring:
LGTM.

Issue #96, opened 2016-04-20
Invalid value at breakpoints using GKE.


Jakedavies:
I get the following in my logs using the node cloud debugger in Kubernetes on GKE, I have the cloud platform enabled for the cluster. 

This occurs after I set a breakpoint in the application, and then hit that piece of code.

[...]

PR #98, opened 2016-04-20
Ensure breakpoint variable's names are strings

ofrobots:
In Node.js 4+, property mirror's names can be numbers in the case of array indices. This caused the controller API to reject breakpoint update messages we sent.

Fixes: #96.

### DIFF
--- a/lib/state.js
+++ b/lib/state.js
@@ -442,17 +442,18 @@ StateResolver.prototype.getMirrorProperties_ = function(mirror) {
 };
 
 StateResolver.prototype.resolveMirrorProperty_ = function(property) {
+  var name = String(property.name);
   if (property.isNative()) {
     return {
-      name: property.name(),
+      name: name,
       varTableIndex: NATIVE_PROPERTY_MESSAGE_INDEX
     };
   }
   if (property.hasGetter()) {
     return {
-      name: property.name(),
+      name: name,
       varTableIndex: GETTER_MESSAGE_INDEX
     };
   }
-  return this.resolveVariable_(property.name(), property.value());
+  return this.resolveVariable_(name, property.value());
 };
--- a/lib/state.js
+++ b/lib/state.js
@@ -442,7 +442,7 @@ StateResolver.prototype.getMirrorProperties_ = function(mirror) {
 };
 
 StateResolver.prototype.resolveMirrorProperty_ = function(property) {
-  var name = String(property.name);
+  var name = String(property.name());
   if (property.isNative()) {
     return {
       name: name,
--- a/test/test-v8debugapi.js
+++ b/test/test-v8debugapi.js
@@ -1,7 +1,7 @@
 /*1* KEEP THIS CODE AT THE TOP TO AVOID LINE NUMBER CHANGES */
 /*2*/'use strict';
 /*3*/function foo(n) {
-/*4*/  return n+42;
+/*4*/  var A = new Array(3); return n+42+A[0];
 /*5*/}
 /*6*/function getterObject() {
 /*7*/  var hasGetter = { _a: 5, get a() { return this._a; }, b: 'hello world' };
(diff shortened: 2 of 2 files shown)


ofrobots:
@matthewloring PTAL.

lib/state.js
```diff
@@ -442,17 +442,18 @@ StateResolver.prototype.getMirrorProperties_ = function(mirror) {
 };
 
 StateResolver.prototype.resolveMirrorProperty_ = function(property) {
+  var name = String(property.name);
```
matthewloring: Is name a function or string? or did it change recently?

ofrobots: typo :(

ofrobots:
@matthewloring  Ready for another look.

test/test-v8debugapi.js
```diff
@@ -43,6 +45,67 @@ function stateIsClean(api) {
...
+    assert.equal(typeof variable.value, 'string');
+  }
+  if (variable.type) {
+    assert.equal(typeof variable.type, 'string');
+  }
+  if (variable.members) {
+    variable.members.forEach(validateVariable);
+  }
+  if (variable.varTableIndex) {
+    assert.ok(Number.isInteger(variable.varTableIndex) &&
```
matthewloring: Are these checks exercised by the included tests?

ofrobots: These checks are certainly exercised, with the exception of `type`. We do not populate the `type` property of a variable. I don't think there is harm on asserting on what we want the type to be if we decide to start populating it in the future.

matthewloring: Specifically, I'm curious about the choice of value for MAX_INT. Is 2^31-1 the upper bound based on proto definition?

ofrobots: Yes: https://cloud.google.com/debugger/api/reference/rpc/google.devtools.clouddebugger.v2#google.devtools.clouddebugger.v2.SourceLocation

test/test-v8debugapi.js
```diff
@@ -28,6 +28,8 @@ var breakpointInFoo = {
   location: { path: 'test-v8debugapi.js', line: 4 }
 };
 
+var MAX_INT = 2147483647;
```
matthewloring: Maybe we could rename or comment this since it is not actually the maximum integer in JS.

lib/v8debugapi.js
```diff
@@ -196,7 +196,12 @@ module.exports.create = function(logger_, config_, fileStats_) {
       var listener = onBreakpointHit.bind(
           null, breakpoint, function(err) {
             delete listeners[num];
-            callback(err);
+            // This method is called from the debug event listener, which
+            // swallows all exception. We defer the callback to make sure the
+            // user errors aren't silenced.
+            setImmediate(function() {
```
matthewloring: Is this change landing in this PR or in #97 ?

ofrobots: In #97.

test/test-v8debugapi.js
ofrobots: Added comment.

test/test-v8debugapi.js
```diff
@@ -43,6 +45,67 @@ function stateIsClean(api) {
   return true;
 }
 
+function validateVariable(variable) {
```
matthewloring: Earlier you mentioned moving this validation into the source as opposed to the tests. Is there a reason to hold off on doing that?

ofrobots: I think it would need slightly more scrutiny as would want to reason about the backoff strategy. Complete the breakpoint with error? Disable the agent?

matthewloring: Ok, we can consider this later. Especially since this validates data we get from V8 as opposed to user inputs.

matthewloring:
LGTM.

test/test-v8debugapi.js
ofrobots: Opened #102 to track this.

PR #101, opened 2016-04-21
Reject breakpoints with unsupported actions

matthewloring:


lib/debuglet.js
```diff
@@ -215,7 +215,14 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
 
         default:
           that.logger_.info('\t' + response.statusCode + ' completed.');
-          that.updateActiveBreakpoints_(body.breakpoints || []);
+          var bps = (body.breakpoints || []).filter(function(bp) {
+            if (bp.action !== 'CAPTURE') {
+              that.logger_.warn('Ignoring breakpoint with non-capture action');
```
ofrobots: include `bp.action` in the message.

matthewloring:
@ofrobots PTAL.

ofrobots:
This should reject the breakpoint to the API with an error status instead of ignoring.

matthewloring:

lib/apiclasses.js
```diff
@@ -35,6 +35,7 @@ function StatusMessage(refersTo, description, isError) {
 /** @const */ StatusMessage.UNSPECIFIED = 'UNSPECIFIED';
 /** @const */ StatusMessage.BREAKPOINT_SOURCE_LOCATION =
                 'BREAKPOINT_SOURCE_LOCATION';
+/** @const */ StatusMessage.BREAKPOINT_ACTION = 'BREAKPOINT_ACTION';
```
ofrobots: StatusMessage.refersTo is a message schema defined by the protobuf. You cannot add a new type to enumeration. The API is going to reject this. (Is there a test?).

See documentation here of the schema here:

https://cloud.google.com/debugger/api/reference/rpc/google.devtools.clouddebugger.v2#google.devtools.clouddebugger.v2.StatusMessage

lib/debuglet.js
```diff
@@ -215,7 +217,21 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
 
         default:
           that.logger_.info('\t' + response.statusCode + ' completed.');
-          that.updateActiveBreakpoints_(body.breakpoints || []);
+          var bps = (body.breakpoints || []).filter(function(bp) {
+            if (bp.action !== 'CAPTURE') {
+              that.logger_.warn('Found breakpoint with invalid action:', bp.action);
+              bp.status = new StatusMessage(StatusMessage.BREAKPOINT_ACTION,
+                BREAKPOINT_ACTION_MESSAGE, true);
+              that.debugletApi_.updateBreakpoint(bp, function(err/*, body*/) {
```
ofrobots: Would it be possible to use `completeBreakpoint_` instead? It would be nice if that was the only place in this file that called `debugletApi_.updateBreakpoint`. Do this only if doesn't require bending backwards.

lib/apiclasses.js
matthewloring: Added a comment for the future and removed the new status message. Once we have the proto inside our project we can avoid these problems.

lib/debuglet.js
matthewloring: `completeBreakpoint_` updates our internal data structures for tracking breakpoints and removes the breakpoint from the v8 debug object. Breakpoints with invalid actions can be updated without ever being entered into the breakpoint map or being reported to V8. We could alternatively provide a separate path through these functions for breakpoints that were never registered but that felt like unnecessary complexity.

lib/debuglet.js
```diff
@@ -215,7 +217,21 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
 
         default:
           that.logger_.info('\t' + response.statusCode + ' completed.');
-          that.updateActiveBreakpoints_(body.breakpoints || []);
+          var bps = (body.breakpoints || []).filter(function(bp) {
+            if (bp.action !== 'CAPTURE') {
```
ofrobots: Let's verify that `bp.action` is always/presently-being provided by the API. The proto seems to suggest this is optional.

lib/debuglet.js
```diff
@@ -215,7 +217,21 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
...
+              that.logger_.warn('Found breakpoint with invalid action:', bp.action);
```
ofrobots: Let's extract this code into a new function, perhaps `rejectBreakpoint_`.

matthewloring: done.

lib/debuglet.js
matthewloring: Good thought, it is optional. The code has been updated.

ofrobots:
The commit message needs to be change to reflect what the change actually does.

lib/debuglet.js
```diff
@@ -215,7 +217,18 @@ Debuglet.prototype.scheduleBreakpointFetch_ = function(seconds) {
 
         default:
           that.logger_.info('\t' + response.statusCode + ' completed.');
-          that.updateActiveBreakpoints_(body.breakpoints || []);
+          var bps = (body.breakpoints || []).filter(function(bp) {
+            var action = bp.action || 'CAPTURE';
+            if (action !== 'CAPTURE') {
+              that.logger_.warn('Found breakpoint with invalid action:', action);
```
ofrobots: You could get more of this code pulled into rejectBreakpoint. Sorry I wasn't clearer earlier. I think there should be minimal logic here. Check action; call reject and return false.

ofrobots:
LGTM w/ comment.

lib/debuglet.js
matthewloring: I think `rejectBreakpoint` should be fully general. Currently `completeBreakpoint` also relies on the status being set outside the function since the status requires information about the call site.

PR #103, opened 2016-04-21
Add projectid label/omit default module label

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
If we are doing this, we should also add the minorversion, module and version labels for AppEngine apps.

ofrobots:
My bad, we already had those.

I think the object layout for layouts might be different from what the API expects. I think the `labels` property is expected to be an array of objects of type `{key: 'something', value: 'something'}`.

matthewloring:
@ofrobots I think the format has been clarified.

ofrobots:
LGTM if the e2e tests pass.

matthewloring:
They do.

PR #107, opened 2016-04-26
Report unique ids when not on GCP

matthewloring:
Previously, the debuglet would report a unique id of `undefined` which is not unique.


matthewloring:
@ofrobots PTAL.

ofrobots:
lgtm.

PR #114, opened 2016-05-11
Attempt to fix travis encrypted keyfiles

matthewloring:
This problem only manifests when testing master so it cannot be checked
with pull requests. It appears that the cluster module is interfering
with travis' file decryption and I believe this change will help.


matthewloring:
@ofrobots PTAL.

PR #116, opened 2016-05-11
Add support for log points

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
What are the semantics when the expression is not convertible to the log format? For example in the test case if you had `n` being a String but the format was `%d`? Do the semantics of `console.log` match what the Debugger expects?

matthewloring:
Good catch. I've updated the formatting to match what the service expects.

test/standalone/test-debuglet.js
```diff
@@ -280,5 +280,15 @@ describe(__filename, function(){
...
     });
   });
+
+  describe('format', function() {
+    it('should be correct', function() {
+      assert.deepEqual(Debuglet.format('hi', [5]), 'hi');
+      assert.deepEqual(Debuglet.format('hi $0', [5]), 'hi 5');
+      assert.deepEqual(Debuglet.format('hi $0 $1', [5, 'there']), 'hi 5 there');
+      assert.deepEqual(Debuglet.format('hi $0 $1', [5]), 'hi 5 $1');
+      assert.deepEqual(Debuglet.format('hi $0 $1 $0', [5]), 'hi 5 $1 5');
```
ofrobots: Some more tests to add: "$$", "$$0", "$00", "hi$01". Also

'hi $0', ['$1', 'foo'] should result in 'hi $1'.

ofrobots:
LGTM if you want to address my comment if addressed in a follow-on.

matthewloring:
That was a good catch. I've uploaded a fix but would welcome a simpler approach if you see one.

test/standalone/test-debuglet.js
matthewloring: Done.

ofrobots: Need another test. "hi $11", passing an argument list with alteat 12 elements.

matthewloring: Done.

lib/debuglet.js
```diff
@@ -434,3 +441,53 @@ Debuglet.mapSubtract = function mapSubtract(A, B) {
   return removed;
 };
 
+/**
+ * Formats the message base with placeholders `$0`, `$1`, etc
+ * by substituting the provided expressions. If more expressions
+ * are given than placeholders extra expressions are dropped.
+ */
+Debuglet.format = function(base, exprs) {
```
ofrobots: I think this code would be *much* simpler if you wrote it as a simple nested loop scanning the input from left to right – no need to tokenize.

ofrobots: Although, if this works, I am not too motivated to make you change it. I just need to convince myself that it works.

lib/debuglet.js
```diff
@@ -341,6 +341,13 @@ Debuglet.prototype.addBreakpoint_ = function(breakpoint, cb) {
...
 
       that.logger_.info('Breakpoint hit!: ' + breakpoint.id);
+      if (breakpoint.action === 'LOG') {
+        var message = Debuglet.format(breakpoint.logMessageFormat,
+          breakpoint.evaluatedExpressions.map(function(v) {
+            return v.value;
+          }));
+        console.log(message);
+      }
       that.completeBreakpoint_(breakpoint);
```
ofrobots: Should log points be one-shot? Are they supposed to be completed on the server and deleted? I would expect they would stick around for some amount of time.

lib/debuglet.js
ofrobots: Okay, I think this works.

ofrobots:
LGTM. Suggested a simpler approach, but I would leave it to you if you want to implement it.

matthewloring:
@ofrobots Ready for another look. Log points now persist until timeout and do not resolve additional locals/arguments.

ofrobots:
LGTM

matthewloring:
@ofrobots Throttling is implemented. Cutting this off here, if anything else comes up it'll be a follow on. PTAL.

ofrobots:
I think we should institute a process where we squash the commits at landing. On this PR, I have to go through the full PR to figure out what you have changed on each iteration. This is not very productive.

matthewloring:
I can do this in the future. Here is a diff of the change since last LGTM:

```
diff --git a/lib/debuglet.js b/lib/debuglet.js
index f2751d4..065e27d 100644
--- a/lib/debuglet.js
+++ b/lib/debuglet.js
@@ -333,7 +333,7 @@ Debuglet.prototype.addBreakpoint_ = function(breakpoint, cb) {
     that.logger_.info('\tsuccessfully added breakpoint  ' + breakpoint.id);
     that.activeBreakpointMap_[breakpoint.id] = breakpoint;
 
-    that.v8debug_.wait(breakpoint, function(err) {
+    that.v8debug_.wait(breakpoint, function waitHandler(err) {
       if (err) {
         that.logger_.error(err);
         cb(err);
@@ -347,6 +347,11 @@ Debuglet.prototype.addBreakpoint_ = function(breakpoint, cb) {
             return v.value;
           }));
         console.log(message);
+        if (!that.completedBreakpointMap_[breakpoint.id]) {
+          setTimeout(function() {
+            that.v8debug_.wait(breakpoint, waitHandler);
+          }, 500);
+        }
       } else {
         that.completeBreakpoint_(breakpoint);
       }

diff --git a/test/e2e/test.js b/test/e2e/test.js
index a4b886d..33cf0a8 100644
--- a/test/e2e/test.js
+++ b/test/e2e/test.js
@@ -223,6 +223,9 @@ function runTest() {
         });
         assert.ok(arg, 'should find the n argument');
         assert.strictEqual(arg.value, '10');
+        console.log('-- checking log point was hit again');
+        console.log(transcript.split('n is: 10').length)
+        assert.ok(transcript.split('n is: 10').length > 4);
         console.log('Test passed');
         process.exit(0);
       })
```

ofrobots:
lgtm.

PR #117, opened 2016-05-12
Improve summarization of breakpoint capture

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/state.js
```diff
@@ -437,9 +454,8 @@ StateResolver.prototype.resolveMirrorFast_ = function(mirror) {
 
 StateResolver.prototype.getMirrorProperties_ = function(mirror) {
   var numProperties = this.config_.capture.maxProperties;
-  var namedProperties = mirror.properties(1, numProperties);
-  var indexedProperties = mirror.properties(2, numProperties);
-  return namedProperties.concat(indexedProperties);
+  var properties = mirror.properties();
```
ofrobots: Does this work with all node versions? I am curious why the original code is the way it is.

ofrobots:
LGTM if the .properties() call isn't the cause for the segfault on node 0.12.

lib/state.js
matthewloring: This does work as it is only executed on newer versions of node already. The culprit was the call to  `propertyNames` above. Fixed.

matthewloring:

ofrobots:
LGTM once the CI is green.

PR #120, opened 2016-05-16
Correct log formatting for non-primitives

matthewloring:
We were not correctly resolving non-primitive types for log points. This
change prevents a full capture from occuring and gives correct logs for
objects and arrays.

### DIFF
--- a/lib/v8debugapi.js
+++ b/lib/v8debugapi.js
@@ -23,7 +23,6 @@
 /** @const */ var state = require('./state.js');
 /** @const */ var logModule = require('@google/cloud-diagnostics-common').logger;
 /** @const */ var apiclasses = require('./apiclasses.js');
-/** @const */ var _ = require('lodash');
 /** @const */ var StatusMessage = apiclasses.StatusMessage;
 
 /** @const */ var messages = {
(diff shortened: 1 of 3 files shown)


matthewloring:
@ofrobots PTAL.

lib/v8debugapi.js
```diff
@@ -485,18 +484,20 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
-    var localConfig;
     if (breakpoint.action === 'LOG') {
-      localConfig = { capture: { maxFrames: 0 } };
-      _.defaultsDeep(localConfig, config);
+      var frame = execState.frame(0);
+      var evaluatedExpressions = [];
+      breakpoint.expressions.forEach(function(exp) {
+        evaluatedExpressions.push(state.evaluate(exp, frame).mirror.value());
+      });
+      breakpoint.evaluatedExpressions = evaluatedExpressions;
```
ofrobots: I think you will need to omit the `evaluatedExpressions` loop above, or perhaps, integrate this code in the loop above. At any rate, looking at the code, I think the expectation for log points would be that we do not re-order the list of evaluated expression. The call to splice above changes this?

lib/v8debugapi.js
```diff
@@ -485,18 +484,20 @@ module.exports.create = function(logger_, config_, fileStats_) {
         }
       }
     }
-    var localConfig;
     if (breakpoint.action === 'LOG') {
-      localConfig = { capture: { maxFrames: 0 } };
-      _.defaultsDeep(localConfig, config);
+      var frame = execState.frame(0);
+      var evaluatedExpressions = [];
+      breakpoint.expressions.forEach(function(exp) {
```
ofrobots: Could this block be replaced by `breakpoint.expressions = breakpoint.expressions.map( ... );`

ofrobots:
The title of the PR, commit doesn't seem accurate anymore? It is not just an optimization

lib/v8debugapi.js

lib/v8debugapi.js
matthewloring: Added a todo.

matthewloring:
done. PTAL.

lib/v8debugapi.js
```diff
@@ -485,18 +484,21 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
     }
-    var localConfig;
     if (breakpoint.action === 'LOG') {
-      localConfig = { capture: { maxFrames: 0 } };
-      _.defaultsDeep(localConfig, config);
+      // TODO: This doesn't work with compiled languages if there is an error
+      // compiling one of the expressions in the loop above.
+      var frame = execState.frame(0);
+      var evaluatedExpressions = breakpoint.expressions.map(function(exp) {
+        return state.evaluate(exp, frame).mirror.value();
```
ofrobots: Object returned by `evaluate` may not necessarily have a `mirror` property.

ofrobots:
LGTM.

PR #122, opened 2016-05-18
Add registration retries

matthewloring:


matthewloring:
@ofrobots PTAL.

lib/debuglet.js
```diff
@@ -142,8 +142,8 @@ Debuglet.prototype.scheduleRegistration_ = function(seconds) {
 
   function onError(err) {
     that.logger_.error('Failed to re-register debuggee: ' + err);
-    that.logger_.error('Disabling gcloud debuglet');
-    that.stop(); // fatal error
+    that.scheduleRegistration_(Math.min((seconds + 1) * 2,
```
ofrobots: Let's just make this simpler and omit the exponential backoff. The default value of 40 is small enough. WDYT?

ofrobots: If you agree, let's rename to name of the config property and drop the `max`.

ofrobots:
LGTM w/ comment.

lib/debuglet.js
matthewloring: I don't think there is a substantial complexity difference so lets keep the backoff.

PR #123, opened 2016-05-18
Configurable throttling for log points

matthewloring:


matthewloring:
@ofrobots PTAL.

matthewloring:
@ofrobots The throttling delay has been reduced to 1s.

ofrobots:
It is unfortunate that we create new listener functions on each log point hit. It would be better to move this logic into v8debugapi.

matthewloring:

lib/v8debugapi.js
```diff
@@ -201,6 +201,40 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+            var currTime = Date.now();
+            if (currTime > timesliceEnd) {
+              logsThisSecond = 0;
+              timesliceEnd = currTime + 1000;
+            }
+            var message = format(breakpoint.logMessageFormat,
+              breakpoint.evaluatedExpressions.map(JSON.stringify));
+            logsThisSecond++;
+            console.log(message);
+            if (!shouldStop()) {
```
ofrobots: nit: reverse the branch and avoid the `!`.

lib/v8debugapi.js
```diff
@@ -201,6 +201,40 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+    /**
+     * @param {Breakpoint} breakpoint
+     * @param {Function} callback
+     */
+    log: function(breakpoint, format, shouldStop) {
+      var logsThisSecond = 0;
+      var timesliceEnd = Date.now() + 1000;
+      var num = breakpoints[breakpoint.id].v8Breakpoint.number();
+      var listener = onBreakpointHit.bind(
+          null, breakpoint, function(err) {
```
cristiancavalli: nit: This is kind of a big function; could this be broken out into a separate function-factory entirely to so that is property isn't as busy?

cristiancavalli: The error callback that is, sry for the ambiguity.

lib/v8debugapi.js

lib/v8debugapi.js
matthewloring: Refactored.

lib/debuglet.js
```diff
@@ -333,27 +333,23 @@ Debuglet.prototype.addBreakpoint_ = function(breakpoint, cb) {
...
-          setTimeout(function() {
-            that.v8debug_.wait(breakpoint, waitHandler);
-          }, 500);
+    if (breakpoint.action === 'LOG') {
+      that.v8debug_.log(breakpoint, Debuglet.format,
+        function() {
+          return that.completedBreakpointMap_[breakpoint.id];
+        });
+    } else {
+      that.v8debug_.wait(breakpoint, function(err) {
```
ofrobots: Any reason to drop the callback function name?

matthewloring: It was previously referenced and this change removes the need to reference it. The name isn't hurting but we generally don't name unreferenced callbacks so I removed it.

ofrobots:
LGTM.

PR #124, opened 2016-05-18
Change language to use snapshot instead of breakpoint

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
LGTM w/ comments (verbal).

PR #125, opened 2016-05-19
Special case array length reporting

matthewloring:


matthewloring:
@ofrobots PTAL.

ofrobots:
As discussed; this looks reasonable, but perhaps a more generic approach that catches all/more built-in object properties would be preferable.

matthewloring:
After discussing this with the V8 team, it seems a more complete solution does not currently exist. We can continue pursuing that approach but I think this patch will address the immediate problem and we are not currently aware of other members in this class of safe to evaluate, native properties (String.length does also fall here but we report strings as literals and do not list their properties so this is not an issue for us).

matthewloring:
It is also worth noting that array length does not show up on the object mirror in Node.js v0.12. This change only prevents the error we were previously reporting in v1+.

lib/state.js
```diff
@@ -460,7 +460,8 @@ StateResolver.prototype.getMirrorProperties_ = function(mirror) {
 
 StateResolver.prototype.resolveMirrorProperty_ = function(property) {
   var name = String(property.name());
-  if (property.isNative()) {
+  var isArrayLen = property.mirror_.isArray() && name === 'length';
```
ofrobots: What is the type of `property`. Is it guaranteed to have a valid `mirror_` property?

matthewloring: It should be some subtype of `PropertyMirror` which defines the `mirror_` field: https://github.com/v8/v8/blob/master/src/debug/mirrors.js#L1466.

ofrobots: Add a comment explaining why the special casing of array is necessary.

ofrobots:
LGTM w/ nits.

lib/state.js
matthewloring: Done.

PR #133, opened 2016-06-07
Prevent calls to deleted breakpoint listeners

matthewloring:


matthewloring:

test/test-v8debugapi.js
```diff
@@ -371,6 +371,21 @@ describe('v8debugapi', function() {
 
     });
 
+    it('should work with multiply hit breakpoints', function(done) {
```
ofrobots: I don't think this test fails even without your fix above. I think we need a proper regression test for the bug you're fixing. I think you need to add a try-catch around the body of handleDebugEvents, and (somehow) to ensure that it actually never catches anything.

matthewloring: Fixed.

lib/v8debugapi.js
```diff
@@ -189,7 +190,7 @@ module.exports.create = function(logger_, config_, fileStats_) {
       var num = breakpoints[breakpoint.id].v8Breakpoint.number();
       var listener = onBreakpointHit.bind(
           null, breakpoint, function(err) {
-            delete listeners[num];
+            listeners[num].enabled = false;
```
cristiancavalli: Since this is an aync function might it be a good idea to check for the `[num]` property before accessing it in case the array is mutated by a bad commit sometime in the future? Example:
```
listeners[num].enabled = false
```
Check first:
```
if (typeof listeners[num] !== undefined) {
  listeners[num].enabled = false;
} 
```

lib/v8debugapi.js
```diff
@@ -198,7 +199,7 @@ module.exports.create = function(logger_, config_, fileStats_) {
             });
           });
 
-      listeners[num] = listener;
+      listeners[num] = { enabled: true, listener: listener };
```
cristiancavalli: nit: This is not actionable right now, but since I've seen this object signature created in several times throughout the agent might it be a good idea to turn it into a shared class?

e.g.
```JS
function ListenerContainer ( enabled=false, listener=null ) {
  this.enabled_ = enabled;
  this.listener_ = listener;
}

ListenerContainer.setIsEnabled ( enabled=false ) {
  this.enabled_ = enabled;

  return this;
}

ListenerContainer.setListener = function ( listener=null ) {
  this.listener_ = listener;

  return this;
}
```

lib/v8debugapi.js
matthewloring: In general this is a good thing to reason about. In this instance, the listener is invoked synchronously by `handleDebugEvents` which includes a `try-catch` that will handle and log any errors that result from the listener not being present so I think we're ok.

lib/v8debugapi.js
matthewloring: I think since this is only done twice in this file I may leave it as is. However, I'll keep an eye out for opportunities to class-ify things (especially if we can get some type-checking in place 😄) .

ofrobots:
LGTM.

PR #135, opened 2016-06-10
Resume logging after log quota is hit

matthewloring:


matthewloring:

lib/v8debugapi.js
```diff
@@ -205,12 +205,33 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
+          logsThisSecond = 0;
+          timesliceEnd = currTime + 1000;
+        }
+        print(breakpoint.logMessageFormat,
+          breakpoint.evaluatedExpressions.map(JSON.stringify));
+        logsThisSecond++;
+        if (shouldStop()) {
+          delete listeners[num];
+        } else {
+          if (logsThisSecond > config.log.maxLogsPerSecond) {
```
cristiancavalli: Is the ceiling more than or `>=`?

matthewloring: Good catch. I have been thinking about this incorrectly. `>=` makes more sense.

matthewloring: Fixed.

cristiancavalli:
LGTM pending CI

PR #136, opened 2016-06-10
Allow ES6 conditions and watches

ofrobots:
We were previously rejecting ES6 expressions. Switch over acorn to ES6
mode and add a few tests.

@matthewloring,@cristiancavalli PTAL.

### DIFF
--- a/test/test-v8debugapi.js
+++ b/test/test-v8debugapi.js
@@ -303,7 +303,9 @@ describe('v8debugapi', function() {
     '!fib()',
     '1+fib()',
     'x++',
-    '[1, 2, 3, 4, x = 1, x == 1, x === 1]'
+    '[1, 2, 3, 4, x = 1, x == 1, x === 1]',
+    '[0].values()',
+    'new Object()',
   ]);
   conditionTests('valid conditions', function(err) { assert.ifError(err); }, [
     null,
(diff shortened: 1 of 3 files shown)


test/test-v8debugapi.js
```diff
@@ -319,9 +321,39 @@ describe('v8debugapi', function() {
...
-    '[1, 2, 3, 4, x == 1, x === 1, null, undefined]'
+    '[1, 2, 3, 4, x == 1, x === 1, null, undefined]',
+    '[0].values',
+    '[][0]',
+    '[0][' + MAX_INT + ']',
+    '"𠮷".length + (5| "𠮷")',
+    '/ٹوٹ بٹوٹ کے دو مُرغے تھے/',
   ]);
 
+  if (semver.satisfies(process.version, '>=4.0')) {
```
matthewloring: Could we also have a check that unsupported ES6 features are handled gracefully on versions of V8 that do not support them?

ofrobots: The purpose of this test is to ensure that side-effect free ES6 expressions are allowed to be evaluated by the debug agent. We already have tests that ensure invalid grammar is gracefully handled. Adding specific ES6 examples that we expect to fail to parse isn't necessarily something that would provide value to these tests. I am not opposed to adding more examples here, but those aren't an objective for this set of tests.

test/test-v8debugapi.js
```diff
@@ -371,6 +403,21 @@ describe('v8debugapi', function() {
 
     });
 
+    it('should work with multiply hit breakpoints', function(done) {
```
matthewloring: This was probably included by mistake?

matthewloring:
LGTM with nit.

test/test-v8debugapi.js
ofrobots: Indeed :smile:. Good catch

ofrobots: Fixed.

PR #154, opened 2016-10-03
Add debuggee name to the re-register log message

### LINKED ISSUE #128: No debugger available on App Engine flex
My `app.js` contains `require('@google/cloud-debug');`
I deployed my Node.js app on App Engine flexible.

The debug view says:
The Cloud Debugger could not find any application to debug:
Discovering debuggable applications

The README says:
"If you are using Google App Engine Managed VMs, you do not have to do any additional configuration."

Why isn't the Cloud Debugger available?

DominicKramer:
If an error occurs when re-registering a debuggee, the name of
the debuggee is reported in the error message to aid in debugging
why the re-registration failed.

fixes #128

### DIFF
--- a/lib/debugletapi.js
+++ b/lib/debugletapi.js
@@ -71,7 +71,7 @@ DebugletApi.prototype.init = function(uid, logger, callback) {
 
   function complete(err, project) {
     if (err) {
-      callback(err);
+      callback(err, project);
       return;
     }
     that.project_ = project;
--- a/lib/debugletapi.js
+++ b/lib/debugletapi.js
@@ -83,7 +83,7 @@ DebugletApi.prototype.init = function(uid, logger, callback) {
         logger.warn('Malformed source-context.json file.');
         // But we keep on going.
       }
-      callback(null);
+      callback(null, project);
     });
   }
 
--- a/test/test-debugletapi.js
+++ b/test/test-debugletapi.js
@@ -53,18 +53,12 @@ describe('Debuglet API', function() {
   });
 
   it('should acquire the project number during init', function(done) {
-    var PROJECT = 'project123';
-    var oldProjNum = utils.getProjectNumber;
-    utils.getProjectNumber = function(callback) {
-      callback(null, PROJECT);
-    };
     debugletapi.init('uid123', { warn: function() {} }, function(err, project) {
       assert(!err);
       // make sure init() invokes the callback with the correct project name
-      assert.equal(project, PROJECT);
+      assert.equal(project, 'project123');
       // make sure the debugletapi is properly storing the project name
       assert.equal(debugletapi.project_, project);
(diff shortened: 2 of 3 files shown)


DominicKramer:

matthewloring:
Could we update https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/test/test-debugletapi.js to check the projectId passed in the callback to `init`. Right now the test validates these using a hack to check for outgoing warnings and using the callback will be cleaner. Also, we want the test to show accurate usage of the api so that people who look there as a source of documentation will see the complete picture.

test/test-debugletapi.js
```diff
@@ -53,7 +53,7 @@ describe('Debuglet API', function() {
   });
 
   it('should acquire the project number during init', function(done) {
-    debugletapi.init('uid123', { warn: function() {} }, function(err) {
+    debugletapi.init('uid123', { warn: function() {} }, function(err, project) {
       assert(!err);
```
matthewloring: Currently this tests whether the project number was retrieved successfully by checking for a lack of errors. It will provide more confidence if we add an assert checking the value of `project` as well now that it is available.

test/test-debugletapi.js
```diff
@@ -53,8 +53,9 @@ describe('Debuglet API', function() {
...
+      assert.equal(debugletapi.project_, project);
```
matthewloring: This test will still pass if the project is not set correctly. Maybe we could compare project to the expected value. It looks like it should be `'project123'` from https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/154/files#diff-8cd5c99c65934398b26d2eb9e3bfc2ddL29

lib/debugletapi.js
```diff
@@ -83,7 +83,7 @@ DebugletApi.prototype.init = function(uid, logger, callback) {
         logger.warn('Malformed source-context.json file.');
         // But we keep on going.
       }
-      callback(null);
+      callback(null, project);
```
ofrobots: There are paths in this function where `callback` is called without the `project` parameter – those are actually the cases where we *need* to know the project number.

Matt's suggestion about the test below would catch this.

DominicKramer:

test/test-debugletapi.js
DominicKramer: This has been fixed.

matthewloring:
LGTM

test/test-debugletapi.js
DominicKramer: Fixed

lib/debugletapi.js
DominicKramer: Now the project name is passed to the callback in the error case as well as the non-error case.

PR #157, opened 2016-10-06
Avoid doubly expiring breakpoints

matthewloring:


matthewloring:
@GoogleCloudPlatform/node-team PTAL.

test/standalone/test-debuglet.js
```diff
@@ -311,6 +311,65 @@ describe(__filename, function(){
     debuglet.start();
   });
 
+  // This test catches regressions in a bug where the agent would
+  // re-schedule and already expired breakpoint to expire if the
```
DominicKramer: You accidentally wrote 'and' instead of 'an'.

DominicKramer:
LGTM

test/standalone/test-debuglet.js
matthewloring: Thanks! Good catch.

PR #158, opened 2016-10-06
Reduce flakiness caused by short timeout

matthewloring:


matthewloring:
@GoogleCloudPlatform/node-team PTAL.

matthewloring:
We won't know for sure whether this fixes the flake but I've given it ~10 travis runs without a failure.

Issue #153, opened 2016-10-02
Off-by-one-line error in transpiled breakpoint


ofrobots:
app.ts

``` typescript
import * as debug from '@google/cloud-debug'
import * as express from 'express';
var app = express();
app.get('/', (req, res) => {
  res.status(200).send(`express + ${debug}`);
});
app.get('/hello/:name', (req, res) => {
  var msg : string = `Hello ${req.params.name}`;
  res.status(200).send(msg);
});

app.listen(process.env.PORT || 8080, function() {
  console.log('App started');
});

export var App = app;
```

Compile with `tsc --sourceMap app.ts`, producing app.js:

``` js
"use strict";
var debug = require('@google/cloud-debug');
var express = require('express');
var app = express();
app.get('/', function (req, res) {
    res.status(200).send("express + " + debug);
});
app.get('/hello/:name', function (req, res) {
    var msg = "Hello " + req.params.name;
    res.status(200).send(msg);
});
app.listen(process.env.PORT || 8080, function () {
    console.log('App started');
});
exports.App = app;
//# sourceMappingURL=app.js.map
```

Set breakpoint at app.ts:9. This corresponds to app.js:10. We, however, resolve this to app.js:9. `msg` is `undefined`.

<img width="1120" alt="screen shot 2016-10-02 at 1 10 36 pm" src="https://cloud.githubusercontent.com/assets/79017/19024359/dd03d3b0-88b6-11e6-9e92-e4d0fb762cde.png">

DominicKramer:
Fixed with PR [#159](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/159).

PR #159, opened 2016-10-14
Add improved support for transpiled code

DominicKramer:
A Sourcemapper class was introduced that provides improved support for transpiled code and has been tested with Babel, Coffeescript, and Typescript.  

In particular, `v8debugapi.js` uses a Sourcemapper to understand the connections between a source file and the transpiled code associated with that source file (if that source file has transpiled code associated with it).  For this to work, the transpilation process used to generate the transpiled code needs to also generate sourcemap files.

This pull request also fixes bug #153 "Off-by-one-line error in transpiled breakpoint".


DominicKramer:

lib/polyfill.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
Splaktar: Perhaps these should be 2016 now?

lib/sourcemapper.js
```diff
@@ -0,0 +1,173 @@
...
+ * 
+ * If there is no such mapping, it could be because the input file is not
+ * the input to a transpilation process or it is the input to a transpilation
+ * process but its corresponding .map file was not given to the constructor
+ * of this mapper.
+ * 
+ * @param {string} inputPath The absolute path to an input file that could
+ *  possibly be the input to a transpilation process
+ */
+Sourcemapper.prototype.hasMappingInfo = function(inputPath) {
```
matthewloring: We may have to relax the absolute path assumption here. If I understand correctly, the input path will be coming from the debug service which may not which may not exactly match the paths observed on the deployed VM.

DominicKramer: The `Sourcemapper` has been updated so that the paths provided to its constructor are paths relative to the process's current working directory instead of absolute paths.  This seems to be more general than using the project's working directory, and it seems that the `node` process needs to be run from within the project's directory so that the `package.json` file can be located.  Let me know if this is not the case and/or you see any problems with this approach.

lib/polyfill.js
DominicKramer: This file is no longer needed and has been deleted, but I updated the copyright notice in the other new files that this pull request creates.

test/standalone/test-sourcemapper.js
```diff
@@ -0,0 +1,417 @@
...
+  [15, 33],
+  [17, 36],
+  [19, 38],
+  [20, 40],
+  [21, 44]
+]);
+
+// ---------------------- supplemental information ------------------------
+
+/*
```
DominicKramer: I've made a new `sourcemapper` directory under `fixtures` to hold the input, output, build, and sourcemap files.

DominicKramer:

lib/v8debugapi.js
```diff
@@ -108,21 +123,22 @@ module.exports.create = function(logger_, config_, fileStats_) {
...
-          line: breakpoint.location.line,
-          column: 0 // unlike lines, columns are 0-indexed for source-map ಠ_ಠ.
-        };
-        compile = getBreakpointCompiler(breakpoint);
+        setInternal(breakpoint, null /* mapInfo */, null /* compile */, cb);
+      }
+      else {
+        var line = breakpoint.location.line,
+            column = 0,
+            mapInfo = sourceMapper.mappingInfo(baseScriptPath, line, column);
```
ofrobots: style nit: use multiple var statements to be consistent with the rest of the project.

lib/scanner.js
```diff
@@ -89,7 +89,7 @@ function computeStats(fileList, shouldHash, callback) {
  * @param {!string} baseDir top-level directory to scan
  * @param {!function(?Error, Array<string>)} callback error-back callback
  */
-function findJSFiles(baseDir, callback) {
+function findFiles(baseDir, regex, callback) {
```
DominicKramer: Fixed.

DominicKramer:
@GoogleCloudPlatform/node-team PTAL.  There is a stacktrace printed out when the unit tests are run, but the tests still pass.  I will look into this, but I wanted to get feedback on the current code.

lib/sourcemapper.js
```diff
@@ -0,0 +1,253 @@
...
+  * properly get the correct mapping information.
+  * 
+  * In particular, the generatedPositionFor() alone doesn't appear to 
+  * give the correct mapping information.
+  */
+  var mappedPos = allPos && allPos.length > 0 ?
+    Array.prototype.reduce.call(allPos,
+      function(accumulator, value, index, arr) {
+          return value.line < accumulator.line ? value : accumulator;
+      }) : consumer.generatedPositionFor(sourcePos);
```
matthewloring: In what case will `allGeneratedPositionsFor` return nothing but `generatedPositionFor` return something?

lib/sourcemapper.js
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * @param {Array.<string>} sourcemapPaths An array of paths to .map sourcemap
+ *  files that should be processed.  The paths should be relative to the 
+ *  current process's current working directory
+ * @param {Logger} logger A logger that reports errors that occurred while
+ *  processing the given sourcemap files
+ * @constructor
+ */
+function Sourcemapper(sourcemapPaths, logger) {
```
matthewloring: Reading through this again, could we simplify things by having a `getSourceMapper` function  which does the asynchronous initialization and then passes off the source mapper instance to a provided callback? This pattern is already done for the scanner and debugapi. Adding a little bit more time on initialization should be fine. What do others think?

ofrobots: I concur. Getting rid of the `waitingCallQueue_` would be a nice simplification.

lib/sourcemapper.js
```diff
@@ -0,0 +1,253 @@
...
+  async.parallel(callList, function(err) {
+    if (err){
+      logger.error('An error occurred while processing the sourcemap files' +
+                   err);
+    }
+
+    that.api_ = new SourcemapperApi(that.infoMap_);
+
+    _.forEach(that.waitingCallQueue_, function(callback){
+      process.nextTick(function() {
```
ofrobots: I am going to assume that this is going to be removed, but in general, prefer `setImmediate` over `nextTick`.

DominicKramer:

I've updated the code to only use callbacks.  In particular, `debuglet.js` now uses `SourceMapper.create()` to create a `SourceMapper` that is provided through a callback.  That `SourceMapper` is then supplied to `v8debugapi.create()`.

lib/sourcemapper.js
```diff
@@ -0,0 +1,209 @@
...
+function create(sourcemapPaths, callback) {
+  var mapper = new SourceMapper();
+  var callList = Array.prototype.slice.call(sourcemapPaths)
+                 .map(function(path) {
+                   return function(callback) {
+                     processSourcemap(mapper.infoMap_, path, callback);
+                   };
+                 });
+
+  async.parallel(callList, function(err) {
```
ofrobots: Perhaps use `async.parallelLimit` to ensure we don't IO-bomb the app startup if there are too many source map paths? I don't think there is a significant urgency in our module to finish the source map process ASAP.

lib/sourcemapper.js
```diff
@@ -0,0 +1,209 @@
...
+ *  SourceMapConsumer objects that are used to calculate mapping information
+ * @param {string} mapPath The path to the sourcemap file to process.  The
+ *  path should be relative to the process's current working directory
+ * @private
+ */
+function processSourcemap(infoMap, mapPath, callback) {
+  // this handles the case when the path is undefined, null, or
+  // the empty string
+  if (!mapPath || !_.endsWith(mapPath, MAP_EXT)){
+    return callback(new Error('The path ' + mapPath +
```
ofrobots: [Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony). This callback is supposed to be async.

lib/sourcemapper.js
```diff
@@ -56,25 +56,31 @@ function processSourcemap(infoMap, mapPath, callback) {
...
+    });
   }
   mapPath = path.normalize(mapPath);
 
   fs.readFile(mapPath, 'utf8',
     function(err, data) {
       if (err){
-        return callback(new Error('Could not read sourcemap file ' + mapPath +
-                        ': ' + err));
+        return setImmediate(function() {
```
ofrobots: It is not necessary to use `setImmediate` here as this is already async (called after `readfile`).

lib/sourcemapper.js
```diff
@@ -56,25 +56,31 @@ function processSourcemap(infoMap, mapPath, callback) {
...
       }
 
       var consumer;
       try {
         consumer = new sourceMap.SourceMapConsumer(data);
       }
       catch(e) {
-        return callback(new Error('An error occurred while reading the '+
-                        'sourcemap file ' + mapPath + ': ' + e));
+        return setImmediate(function() {
```
ofrobots: ditto.

lib/sourcemapper.js
```diff
@@ -100,8 +106,10 @@ function processSourcemap(infoMap, mapPath, callback) {
         });
 
       if (sources.length === 0) {
-        return callback(new Error('No sources listed in the sourcemap file ' +
-          mapPath));
+        return setImmediate(function() {
```
ofrobots: ditto.

lib/sourcemapper.js
```diff
@@ -112,7 +120,9 @@ function processSourcemap(infoMap, mapPath, callback) {
         });
       });
 
-      callback(null);
+      setImmediate(function() {
```
ofrobots: ditto.

lib/sourcemapper.js
DominicKramer: Fixed

lib/sourcemapper.js
DominicKramer: Fixed

lib/sourcemapper.js
DominicKramer: Fixed

lib/sourcemapper.js
DominicKramer: Fixed

PR #162, opened 2016-10-20
Initial commit to prioritize capturing expressions

DominicKramer:
In particular, now if a captured value has a `length` attribute,
the value of that attribute is shown in the status field.

Also, the length of a watched expression is now bounded by
`config.maxDataSize` instead of `config.maxProperties`.

Unit tests still need to be addressed, and more fine grained
control of the total length of the data captured needs to be
added.


DominicKramer:
@GoogleCloudPlatform/node-team I'm still working on this, but I opened this pull request to get early feedback so that if my changes will have a negative impact that I am not aware of, I can fix it sooner rather than later.

lib/state.js
```diff
@@ -157,7 +164,7 @@ StateResolver.prototype.capture_ = function() {
                                     result.error, true)
         };
       } else {
-        evaluated = that.resolveVariable_(expression, result.mirror);
+        evaluated = that.resolveVariable_(expression, result.mirror, true);
```
matthewloring: This boolean seems to propagate parallel information with the `isEvaluated` property above. I would be in favor of using the boolean parameter alone to avoid modifying the V8 object.

DominicKramer:
PTAL

lib/state.js
```diff
@@ -157,21 +162,34 @@ StateResolver.prototype.capture_ = function() {
...
-         (that.totalSize_ < that.config_.capture.maxDataSize || noLimit)) {
+  while (index < that.rawVariableTable_.length) { // NOTE: length changes in loop
     assert(!that.resolvedVariableTable_[index]); // shouldn't have it resolved yet
+    var isEvaluated = evalIndexSet.has(index);
     that.resolvedVariableTable_[index] =
-      that.resolveMirror_(that.rawVariableTable_[index]);
+      that.resolveMirror_(that.rawVariableTable_[index], isEvaluated);
     index++;
+
+    if (!noLimit && that.totalSize_ >= that.config_.capture.maxDataSize) {
```
ofrobots: nit: readability would be improved if you reversed `noLimit` to `haveLimit` above and avoided a `!` here.

lib/state.js
```diff
@@ -468,10 +486,15 @@ StateResolver.prototype.resolveVariable_ = function(name, value) {
...
     var maxProps = this.config_.capture.maxProperties;
-    if (maxProps && maxProps < Object.keys(value.value()).length) {
+    var maxBuffer = this.config_.capture.maxDataSize;
+    var numKeys = Object.keys(value.value()).length;
+
+    if (!isEvaluated && maxProps && maxProps < numKeys) {
       data.status = MESSAGE_TABLE[OBJECT_LIMIT_MESSAGE_INDEX].status;
     }
-
+    else if (isEvaluated && maxBuffer > 0 && maxBuffer < numKeys) {
```
ofrobots: I don't know if this test makes a lot of sense. `maxBuffer` is supposed to be the proportional number of bytes in the buffer we send. `numKeys` and `maxBuffer` don't seem like quantities you can compare.

I think what you really want to figure out is how many properties to encode given that the amount of space *left* in the buffer.

Perhaps we should use a different approach here wherein the `maxDataSize` limits are applied only at the top-level in `capture_`. That is, `resolveVariable_` and `resolveMirror_` wouldn't care about `maxDataSize`. WDYT?

lib/state.js
DominicKramer: This has been fixed to remove the duplicated information without modifying the V8 object.

lib/state.js
DominicKramer: Fixed.

lib/state.js
```diff
@@ -446,7 +464,7 @@ StateResolver.prototype.resolveLocalsList_ = function (frame, args) {
  * @param {String} name The name of the variable.
  * @param {Object} value A v8 debugger representation of a variable value.
  */
-StateResolver.prototype.resolveVariable_ = function(name, value) {
+StateResolver.prototype.resolveVariable_ = function(name, value, isEvaluated) {
```
DominicKramer: Added.

lib/state.js
DominicKramer: Now `maxDataSize` is not being used in comparisons that compare against the number of attributes in an object.  Instead, an evaluated expression is either entirely collected or not entirely collected depending on whether or not the entire contents of that expression would go beyond the limits of `maxDataSize`.

lib/state.js
```diff
@@ -468,10 +485,11 @@ StateResolver.prototype.resolveVariable_ = function(name, value) {
   } else if (value.isObject()) {
     data.varTableIndex = this.getVariableIndex_(value);
     var maxProps = this.config_.capture.maxProperties;
-    if (maxProps && maxProps < Object.keys(value.value()).length) {
+    var numKeys = Object.keys(value.value()).length;
```
cristiancavalli: Prefer _.keys over Object.keys due to API compatibility across Node.JS versions

lib/state.js
```diff
@@ -468,10 +485,11 @@ StateResolver.prototype.resolveVariable_ = function(name, value) {
...
+
+    if (!isEvaluated && maxProps && maxProps < numKeys) {
```
cristiancavalli: `<` or `<=` ?

lib/state.js
```diff
@@ -510,53 +528,68 @@ StateResolver.prototype.storeObjectToVariableTable_ = function(obj) {
...
 
 // A slower implementation of resolveMirror_ which is safe for all node versions
-StateResolver.prototype.resolveMirrorSlow_ = function(mirror) {
+StateResolver.prototype.resolveMirrorSlow_ = function(mirror, isEvaluated) {
   // Instead, let's use Object.keys. This will only get the enumerable
   // properties. The other alternative would be Object.getOwnPropertyNames, but
   // I'm going with the former as that's what util.inspect does.
   var that = this;
+
   var keys = Object.keys(mirror.value());
```
cristiancavalli: Can we retroactively replace this call to `Object.keys` with `_.keys`?

lib/state.js
```diff
@@ -510,53 +528,68 @@ StateResolver.prototype.storeObjectToVariableTable_ = function(obj) {
...
 };
 
 // A faster implementation of resolveMirror_ which segfaults in node <1.6
 //
 // See https://github.com/iojs/io.js/issues/1190.
-StateResolver.prototype.resolveMirrorFast_ = function(mirror) {
-  var members = this.getMirrorProperties_(mirror).map(
-      this.resolveMirrorProperty_.bind(this));
+StateResolver.prototype.resolveMirrorFast_ = function(mirror, isEvaluated) {
+  var resMirrorProp = this.resolveMirrorProperty_.bind(this);
```
cristiancavalli: Should we be using `.bind` here or a closure due to performance concerns across varying Node versions?
cc @ofrobots

test/test-v8debugapi.js
```diff
@@ -727,65 +735,139 @@ describe('v8debugapi', function() {
...
 
-          api.clear(bp);
-          config.capture.maxProperties = oldMax;
-          done();
+            api.clear(bp);
+            config.capture.maxDataSize = oldMaxData;
+            config.capture.maxProperties = oldMaxProps;
+            done();
+          });
+          process.nextTick(function() {foo(2);});
```
cristiancavalli: Prefer use of `setImmediate` over `process.nextTick`?
cc @ofrobots

test/test-v8debugapi.js
```diff
@@ -727,65 +735,139 @@ describe('v8debugapi', function() {
...
+            assert(fooVal.status.description.format.indexOf(
+              'Max data size reached') !== -1);
+            assert(fooVal.status.isError);
+
+            api.clear(bp);
+            config.capture.maxDataSize = oldMaxData;
+            config.capture.maxProperties = oldMaxProps;
+            done();
+          });
+          process.nextTick(function() {foo(2);});
```
cristiancavalli: Prefer use of `setImmediate` over `process.nextTick`?

test/test-v8debugapi.js
```diff
@@ -727,65 +735,139 @@ describe('v8debugapi', function() {
...
 
-          api.clear(bp);
-          config.capture.maxProperties = oldMax;
-          done();
+            api.clear(bp);
+            config.capture.maxDataSize = oldMaxData;
+            config.capture.maxProperties = oldMaxProps;
+            done();
+          });
+          process.nextTick(function() {foo(2);});
```
cristiancavalli: Prefer use of `setImmediate` over `process.nextTick`?

test/test-v8debugapi.js
```diff
@@ -727,65 +735,139 @@ describe('v8debugapi', function() {
...
+            assert(fooVal.status.description.format.indexOf(
+              'Max data size reached') !== -1);
+            assert(fooVal.status.isError);
+
+            api.clear(bp);
+            config.capture.maxDataSize = oldMaxData;
+            config.capture.maxProperties = oldMaxProps;
+            done();
+          });
+          process.nextTick(function() {foo(2);});
```
cristiancavalli: Prefer use of `setImmediate` over `process.nextTick`?

lib/state.js
```diff
@@ -349,7 +364,7 @@ StateResolver.prototype.extractArgumentsList_ = function (frame) {
 StateResolver.prototype.resolveArgumentList_ = function(args) {
   var resolveVariable = this.resolveVariable_.bind(this);
```
cristiancavalli: Change use of `.bind` for closure?

lib/state.js
matthewloring: What are the compatibility concerns with `Object.keys`?

lib/state.js
DominicKramer: I was looking at it from the perspective if the max number of properties was `10`, for example, and an Object has `10` properties, then we shouldn't display an error. So I don't think `<=` is needed.

lib/state.js
DominicKramer: I too am curious about why `_.keys` should be preferred over `Object.keys`.

test/test-v8debugapi.js
DominicKramer: The existing tests used `process.nextTick` rather than `setImmediate`.  Before making a change, I want to make sure the use of `process.nextTick` wasn't a necessity in this particular case.

test/test-v8debugapi.js
DominicKramer: The existing tests used process.nextTick rather than setImmediate. Before making a change, I want to make sure the use of process.nextTick wasn't a necessity in this particular case.

test/test-v8debugapi.js
DominicKramer: The existing tests used process.nextTick rather than setImmediate. Before making a change, I want to make sure the use of process.nextTick wasn't a necessity in this particular case.

test/test-v8debugapi.js
DominicKramer: The existing tests used process.nextTick rather than setImmediate. Before making a change, I want to make sure the use of process.nextTick wasn't a necessity in this particular case.

lib/state.js
DominicKramer: Do you think a closure would be more performant in this situation?

lib/state.js
cristiancavalli: @DominicKramer look at the implementation for lodash keys:
https://github.com/lodash/lodash/blob/4.16.6/lodash.js#L3491

The Object.keys API does not exist ubiquitously

matthewloring: It appears to exist back to at least node 0.8. I think `Object.keys` will be safe here.

test/test-v8debugapi.js
matthewloring: `process.nextTick` was not selected intentionally here. `setImmediate` would be fine.

lib/state.js
ofrobots: Until [recently](http://benediktmeurer.de/2016/01/14/optimizing-bound-functions-further/) bind used to be quite slow. With Node v7.x there should be no difference in performance but earlier versions should have different degrees of performance issues with `bind`.

Having said this, there are other occurrences of `bind` in this file, and I'll let it up to change them in a follow-on change.

test/test-v8debugapi.js
ofrobots: This is a test, so no biggie. If anyone is motivated enough to change things to `setImmediate` it would be better to do it in a follow-on that changes all such in this file.

lib/state.js
DominicKramer: I will leave `bind` for now and will fix all uses of `bind` in a follow-up PR.

lib/state.js
DominicKramer: Ok, I will leave it as `Object.keys`.

test/test-v8debugapi.js
DominicKramer: I will update the tests to use `setImmediate` in a follow-up PR.

cristiancavalli:
LGTM

PR #166, opened 2016-10-27
Update to Specifying Configuration When Starting the Cloud Debugger

DominicKramer:
This PR contains an update where the new `start()` method can be called on the object returned from `require('@google/cloud-debug)` to start the debugger while specifying a configuration.

Additionally, the configuration can now specify custom values for the app's service name and service version.

PR #167, opened 2016-10-28
Add the start() Method and the Ability to Specify the Service Name/Version in the Debug Config

DominicKramer:
This PR contains an update where the new start() method can be called on the object returned from require('@google/cloud-debug) to start the debugger while specifying a configuration.

Additionally, the configuration can now specify custom values for the app's service name and service version.

This PR replaces PR [#166](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/166)


DominicKramer:

This PR replaces PR [#166](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/166), which I closed to avoid confusion.

jmdobry:
@GoogleCloudPlatform/node-team 

Everybody on `@GoogleCloudPlatform/node-team` is watching the cloud-debug-nodejs and cloud-trace-nodejs repositories already, and therefore gets notifications. Is it okay with y'all if we cut back on using `@GoogleCloudPlatform/node-team` to explicitly re-notify everybody?

index.js
```diff
@@ -44,17 +45,41 @@ var initConfig = function() {
...
+  log = logger.create(config.logLevel, '@google/cloud-debug');
+  if (config.enabled) {
+    var debuglet = new Debuglet(config, log);
+    debuglet.start();
+    module.exports.private_ = debuglet;
+    hasStarted = true;
+  }
 }
+
+// exports is populated by the agent
```
matthewloring: I don't believe this comment holds anymore. I think we can remove it.

index.js
```diff
@@ -44,17 +45,41 @@ var initConfig = function() {
...
+  start: start
+};
+
+setTimeout(function() {
+  if (!hasStarted){
+    start();
+    log.error('The cloud-debug agent has been automatically started.  ' +
+      'This action will be deprecated in the future.');
+  }
+}, 5*1000);
```
matthewloring: Could we add a test to ensure that this correctly starts the agent and does not interfere with an already started agent?

index.js
DominicKramer: I have removed it.

index.js
DominicKramer: I've added the tests.

index.js
```diff
@@ -44,17 +45,46 @@ var initConfig = function() {
...
+    debuglet.start();
+    module.exports.private_ = debuglet;
+    started_ = true;
+  }
+}
+
+setTimeout(function() {
+  if (!started_){
+    start();
+    log_.error('The cloud-debug agent has been automatically started.  ' +
```
DominicKramer: I've updated the error message except I state that there is a 1 second delay.

index.js
```diff
@@ -44,17 +45,50 @@ var initConfig = function() {
...
+  }
   return config;
 };
 
-// exports is populated by the agent
-module.exports = {};
-var config = initConfig();
+module.exports = {
+  start: start,
+  hasStarted: hasStarted
```
ofrobots: I believe this is for testing only. Prefer to rename this to `hasStarted_`.

matthewloring:
Could you update the title to reflect that this introduces the `start` function/configuration flow as well?

PR #170, opened 2016-11-01
Add a configuration section to the README

DominicKramer:
This PR adds a configuration section to the README file that should have been added in PR [#167](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/167).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,22 +27,10 @@ Deploy your application, and navigate to the [Stackdriver Debug view][debug-tab]
 
 ## Configuration
 
-First, some configuration options can be specified through environment variables:
-
-* `GCLOUD_DEBUG_DISABLE` specifies whether to enable or disable the agent
-* `GCLOUD_DEBUG_LOGLEVEL` specifies the debug level
-* `GCLOUD_DEBUG_REPO_APP_PATH` specifies the path within your repository to the directory containing the `package.json` for your deployed application
-* `GAE_MODULE_NAME` specifies the service name of the deployed application
-* `GAE_MODULE_VERSION` specifies the version of the deployed application
-
-Next, see [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command as shown below:
+See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command as shown below:
 ```
 require('@google/cloud-debug').start({logLevel: 2});
 ```
-
-Last, you can also provide configuration through a config file.  You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
-
-Note that configuration options specified through environment variables take precedence over options passed to the start command, which take precedence over options specified in a config file.
 
 ## Running on Google Cloud Platform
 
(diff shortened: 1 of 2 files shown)


DominicKramer:
PTAL

README.md
```diff
@@ -20,11 +20,30 @@ This module provides Stackdriver Debugger support for Node.js applications. [Sta
...
 
+## Configuration
+
+First, some configuration options can be specified through environment variables:
+
+* `GCLOUD_DEBUG_DISABLE` specifies whether to enable or disable the agent
+* `GCLOUD_DEBUG_LOGLEVEL` specifies the debug level
+* `GCLOUD_DEBUG_REPO_APP_PATH` specifies the path within your repository to the directory containing the `package.json` for your deployed application
+* `GAE_MODULE_NAME` specifies the service name of the deployed application
+* `GAE_MODULE_VERSION` specifies the version of the deployed application
```
ofrobots: I would prefer to keep the environment variables undocumented for now specially given that we are planning to reverse the precedence in the future (in a semver-major update). It would be best if fewer people were dependent on these and their semantics.

README.md
```diff
@@ -20,11 +20,30 @@ This module provides Stackdriver Debugger support for Node.js applications. [Sta
...
+
+Next, see [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command as shown below:
+```
+require('@google/cloud-debug').start({logLevel: 2});
+```
+
+Last, you can also provide configuration through a config file.  You can start by copying the default config file and modifying it to suit your needs. The `GCLOUD_DIAGNOSTICS_CONFIG` environment variable should point to your configuration file.
+
+Note that configuration options specified through environment variables take precedence over options passed to the start command, which take precedence over options specified in a config file.
```
ofrobots: If we don't document the env. vars, we shouldn't document their precedence.

README.md
```diff
@@ -20,11 +20,18 @@ This module provides Stackdriver Debugger support for Node.js applications. [Sta
...
+# Require and start the agent at the top of your main script (but after '@google/cloud-trace' if you are also using it).
+require('@google/cloud-debug').start();
 ```
 Deploy your application, and navigate to the [Stackdriver Debug view][debug-tab] within the [Google Cloud Console][dev-console] to set snapshots and start debugging.
 
+## Configuration
+
+See [the default configuration](config.js) for a list of possible configuration options. These options can be passed to the agent through the object argument to the start command as shown below:
+```
+require('@google/cloud-debug').start({logLevel: 2});
```
ofrobots: Can we use a serviceContext as an example here? For GCE and GKE we should be suggesting that the users provide the serviceContext configuration.

ofrobots: Other alternatives that might also make more compelling examples. `projectId` or `keyFilename`.

README.md
DominicKramer: Now the environment variables are not documented.

README.md
DominicKramer: Now the environment variables are not documented.

README.md
DominicKramer: I've added `serviceContext` to the example.

DominicKramer: I've moved the Configuration section after all of the 'Running on...' sections.

PR #173, opened 2016-11-11
Correct the status shown if maxDataSize is reached

DominicKramer:
If a watched expression is collected that causes the `config.capture.maxDataSize` limit to be reached, local variables may not be collected, but the status message associated with the local variable would be incorrect.  In particular, if the local was an array, the message would describe that only the first `config.capture.maxProperties` items were captured.  This PR corrects this so that the status message is instead 'Max data size reached'.


DominicKramer:
PTAL

PR #174, opened 2016-11-11
Evaluated expressions respect capture.maxProperties

DominicKramer:
Now whether or not a captured variable is an evaluated
expression, the `config.capture.maxProperties` configuration will
be respected when displaying the contents of the variable.


DominicKramer:
PTAL

PR #175, opened 2016-11-13
improve UX for truncated objects properties

ofrobots:
Change how we indicate that an object with too many properties has been
truncated on our state captures. Instead of a non-error StatusMessage
add an extra member with no value / varTableIndex. This matches how the
Java agent deals with this situation.

Original:
image

Now: 
image

and:
image

R= @DominicKramer @cristiancavalli

### DIFF
--- a/test/test-v8debugapi.js
+++ b/test/test-v8debugapi.js
@@ -660,7 +660,9 @@ describe('v8debugapi', function() {
           assert.equal(procEnv.name, 'process.env');
           var envVal = bp.variableTable[procEnv.varTableIndex];
           envVal.members.forEach(function(member) {
-            assert(bp.variableTable[member.varTableIndex].status.isError);
+            if (member.varTableIndex) {
+               assert(bp.variableTable[member.varTableIndex].status.isError);
+            }
           });
           var hasGetter = bp.evaluatedExpressions[1];
           var getterVal = bp.variableTable[hasGetter.varTableIndex];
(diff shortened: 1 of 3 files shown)


DominicKramer:
Except for the comment I had about what we should display in the case when the `maxDataSize` is reached as well as an object having some properties truncated, it LGTM.

test/test-v8debugapi.js
```diff
@@ -757,10 +759,9 @@ describe('v8debugapi', function() {
           assert.ifError(err);
           var foo = bp.evaluatedExpressions[0];
           var fooVal = bp.variableTable[foo.varTableIndex];
-          assert.equal(fooVal.members.length, 1);
-          assert(foo.status.description.format.indexOf(
-            'Only first') !== -1);
-          assert(!foo.status.isError);
+          // should have 1 element + truncation message.
+          assert.equal(fooVal.members.length, 2);
```
ofrobots: Intended. This PR only changes the UX when objects are truncated due to `maxProperties`. The `maxDataSize` will be handled in a follow on change.

test/test-v8debugapi.js
```diff
@@ -660,7 +660,9 @@ describe('v8debugapi', function() {
           assert.equal(procEnv.name, 'process.env');
           var envVal = bp.variableTable[procEnv.varTableIndex];
           envVal.members.forEach(function(member) {
-            assert(bp.variableTable[member.varTableIndex].status.isError);
+            if (member.varTableIndex) {
```
cristiancavalli: Would this resolve for a member with a `varTableIndex` of `0`?

cristiancavalli:
LGTM with question and passing CI

test/test-v8debugapi.js
ofrobots: Good catch. Fixed.

PR #177, opened 2016-11-18
Now status messages include config values

DominicKramer:
Now if a status message referes to a configuration option the
syntax `configname=configvalue` will be used to specify the
configuration option's name as well as its value.

Also deleted the item at the `DATA_LIMIT_MESSAGE_INDEX` index
in the `MESSAGE_TABLE` since it is no longer needed.


DominicKramer:
PTAL

PR #180, opened 2016-12-07
API changes bring us closer to `google-cloud`

ofrobots:
This change adapts the primary interface of this module to align it
with the API exposed by all the other Google Cloud API Client
Libraries available via the `google-cloud` modules [1].

The primary change is that we return a Debug construtor as the return
value from the module. This, in the future, will provide a Client
Library interface to the Debug APIs. This can be used for programatic
access to list or modify breakpoints, logpoints etc. The constructor,
like other API clients, accepts authentication and other general
configuration.

An instantiated debug object exposes a function `startAgent` that can
be used to activate the process-wide agent that enables the debugging
of an application using the Stackdriver Debug UI in the Google Cloud
console.

Feel free to bike-shed on `startAgent` vs `start`.


package.json
```diff
@@ -70,5 +43,9 @@
     "semver": "^5.1.0",
     "source-map": "^0.5.1",
     "split": "^1.0.0"
+  },
+  "scripts": {
+    "test": "./bin/run-test.sh",
```
matthewloring: If we don't intend to continue using the changelog/publish scripts we should remove them.

package.json
```diff
@@ -52,6 +24,7 @@
     "changelog-maker": "^2.2.2",
     "closure-npc": "*",
```
matthewloring: is it time to drop this?

ofrobots: Yes! but that's not in the scope of this PR.

package.json
ofrobots: Yes! but let's now grow the scope of this PR.

src/index.js
```diff
@@ -56,32 +87,32 @@ var initConfig = function(config_) {
...
-    module.exports.private_ = debuglet;
-    start.wasSuccessful_ = true;
+    this.private_ = debuglet;
   }
-}
+};
+
+module.exports = Debug;
 
-setTimeout(function() {
```
matthewloring: Do we want to pull out this startup logic as part of this change?

package.json
matthewloring: We should preserve all scripts for now then so that we don't have dead code/deps.

ofrobots:
Ready for another look.

src/index.js
```diff
@@ -16,14 +16,45 @@
...
+ * 
+ * This module provides Stackdriver Debugger support for Node.js applications. 
+ * [Stackdriver Debugger](https://cloud.google.com/debug/) is a feature of 
+ * [Google Cloud Platform](https://cloud.google.com/) that lets you debug your
+ * applications in production without stopping or pausing your application.
+ * 
+ * This module provides an agent that lets you automatically enable debugging
+ * without changes to your application. 
+ * 
+ * A low level Client Library for the Debug will also be available through
```
matthewloring: Should this be included?

src/index.js
```diff
@@ -16,14 +16,45 @@
...
+ * this module, but is not present at the moment.
+ * 
+ * @constructor
+ * @alias module:debug
```
matthewloring: What does the `:` syntax mean?

ofrobots: It follows the custom jsdoc parser format used by `google-cloud-node`.

PR #183, opened 2016-12-09
Switch from findit to findit2

ofrobots:
https://www.npmjs.com/package/findit2 seems to fix a few open bugs with
findit, which doesn't seem to be maintained anymore. Furthermore, this
fixes the issue that findit was intermittently failing to find all the
files on windows causing the AppVeyor builds to fail.

### DIFF
--- a/lib/scanner.js
+++ b/lib/scanner.js
@@ -19,7 +19,7 @@ var _ = require('lodash');
 var fs = require('fs');
 var path = require('path');
 var crypto = require('crypto');
-var findit = require('findit');
+var findit = require('findit2');
 var split = require('split');
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


ofrobots:
PTAL. AppVeyor is green again!

ofrobots:
`walker` is indeed more popular – in fact I compared the output from `findit` to `walker` to test if `findit` was buggy. 

However, `findit2` has a compatible API to `findit` and was a minimal change to our codebase.

matthewloring:
Yup, sounds good!

DominicKramer:
I second using `findit2` for the above mentioned minimal changes to our codebase.

PR #188, opened 2016-12-14
listBreakpoint querystring encoding was incorrect

ofrobots:


src/debugletapi.js
```diff
@@ -225,12 +226,13 @@ DebugletApi.prototype.register_ = function(errorMessage, callback) {
...
-  var url = API + '/debuggees/' + encodeURIComponent(that.debuggeeId_) +
-      '/breakpoints';
+  var query = { success_on_timeout: true };
   if (that.nextWaitToken_) {
-    url += '?waitToken=' + encodeURIComponent(that.nextWaitToken_);
+    query.waitToken = that.nextWaitToken;
   }
-  url += '?success_on_timeout=' + encodeURIComponent(true);
+
+  var url = API + '/debuggees/' + encodeURIComponent(that.debuggeeId_) +
```
matthewloring: Are there any tests that exercise the `waitToken` case? They should have been failing while this was being incorrectly encoded.

ofrobots: The e2e test world have been exercising this path. My guess is that the 'success_on_timeout' flag would have gotten dropped by the remote service – this isn't fatal – it would have meant that we would get 404 on timeouts (which would get correctly dealt with), but anyone who was monitoring the error rate of their apps would have seen unnecessary signal.

ofrobots:
Added a bug (internal) to track lack of test coverage for success_on_timeout.

PR #190, opened 2016-12-14
switch to using @google-cloud/common

ofrobots:
This PR starts using the authorized / retry request flow from the
@google-cloud/common library. From an operational perspective This
shouldn't change behaviour significantly – although there are small
differences in how the new library does retries or how it prioritizes
different sources of the projectId & credentials (environment vs.
metadata vs. config).

### DIFF
--- a/test/auth-request.js
+++ b/test/auth-request.js
@@ -1,0 +1,23 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


test/auth-request.js
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+var request = require('request');
+module.exports = function(options, callback) {
+  request(options, function(err, response, body) {
```
matthewloring: :weary:

ofrobots:
Travis fix: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/191

ofrobots:
We also need https://github.com/GoogleCloudPlatform/google-cloud-node/pull/1879 to be able to drop the dependency on diagnostics-common.

ofrobots:
Further, even though this isn't a blocker for development, before we release we would want a resolution on https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1822.

PR #192, opened 2016-12-15
Refactor and rename api classes

ofrobots:
Various refactorings that are essentially functionally equivalent, but align us better to include a Client Library for the Debug API in the future.


src/debuggee.js
```diff
@@ -0,0 +1,110 @@
...
+ *     way of noticing.
+ */
+function Debuggee(projectId, uid, serviceContext, sourceContext, description,
+                  errorMessage) {
+  if (!(this instanceof Debuggee)) {
+    return new Debuggee(projectId, uid, serviceContext, sourceContext,
+                        description, errorMessage);
+  }
+
+  if (typeof projectId !== 'string') {
```
cristiancavalli: Suggestion: since lodash is already included as a dependency use `isString` for brevity of expression

src/debuggee.js
```diff
@@ -0,0 +1,110 @@
...
+    'process.title': process.title,
+    'node version': process.versions.node,
+    'V8 version': process.versions.v8,
+    'agent.name': pjson.name,
+    'agent.version': pjson.version,
+    'projectid': projectId
+  };
+
+  if (serviceContext) {
+    if (typeof serviceContext.service === 'string' &&
```
cristiancavalli: See prior suggestion on use of isString

src/debuggee.js
```diff
@@ -0,0 +1,110 @@
...
+  this.uniquifier = uniquifier;
+  this.description = desc;
+  this.agentVersion = version;
+  this.labels = labels;
+  if (sourceContext) {
+    this.sourceContexts = [sourceContext];
+  }
+}
+
+module.exports = Debuggee;
```
cristiancavalli: Suggestion: add trailing newline

test/test-debuggee.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('should have an error statusMessage with the appropriate arg', function() {
+    var debuggee = new Debuggee('a', 'b', undefined, undefined, undefined,
+                                'Some Error Message');
+    assert.ok(debuggee);
+    assert.ok(debuggee.statusMessage);
+  });
+
+});
```
cristiancavalli: Suggestion: add trailing newline

src/status-message.js
```diff
@@ -41,3 +35,5 @@ function StatusMessage(refersTo, description, isError) {
 /** @const */ StatusMessage.BREAKPOINT_EXPRESSION = 'BREAKPOINT_EXPRESSION';
 /** @const */ StatusMessage.VARIABLE_NAME = 'VARIABLE_NAME';
 /** @const */ StatusMessage.VARIABLE_VALUE = 'VARIABLE_VALUE';
+
+module.exports = StatusMessage;
```
cristiancavalli: Suggestion: add trailing newline

src/debuggee.js
```diff
@@ -0,0 +1,113 @@
...
+    'process.title': process.title,
+    'node version': process.versions.node,
+    'V8 version': process.versions.v8,
+    'agent.name': pjson.name,
+    'agent.version': pjson.version,
+    'projectid': projectId
+  };
+
+  if (serviceContext) {
+    if (typeof serviceContext.service === 'string' &&
```
cristiancavalli: Suggestion: use `isString` for purposes of being concise

src/controller.js
```diff
@@ -126,66 +121,12 @@ DebugletApi.prototype.registerError = function(message) {
...
-    labels: labels,
-    sourceContexts: [that.sourceContext_]
-  };
-
-  if (errorMessage) {
-    debuggee.status = new StatusMessage(StatusMessage.UNSPECIFIED, errorMessage,
-                                        true);
-  }
+  var debuggee = new Debuggee(
+      that.project_, that.uid_,
```
matthewloring: Can these be references to `this`? We could move `that` down closer to the point where it will be observed asynchronously.

src/debuggee.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+function Debuggee(projectId, uid, serviceContext, sourceContext, description,
+                  errorMessage, onGCP) {
+  if (!(this instanceof Debuggee)) {
+    return new Debuggee(projectId, uid, serviceContext, sourceContext,
+                        description, errorMessage, onGCP);
+  }
+
+  if (typeof projectId !== 'string') {
+    throw new Error('projectId must be a string');
```
matthewloring: Do we expect users of our module to call this function directly?

src/index.js
```diff
@@ -113,5 +114,33 @@ Debug.prototype.startAgent = function(config) {
...
 };
 
-module.exports = Debug;
+/**
+ * Attempt to retrieve the project ID from the auth client.
+ *
+ * @param {function} callback - The callback function.
+ */
+Debug.prototype.getProjectId = function(callback) {
+  if (global.GCLOUD_SANDBOX_ENV) {
```
matthewloring: What path makes use of this?

src/index.js
```diff
@@ -113,5 +114,33 @@ Debug.prototype.startAgent = function(config) {
...
+    return;
+  }
+
+  var self = this;
```
matthewloring: Maybe simpler to store the `projectId` here and reference it in the callback as opposed to `this`.

system-test/test-controller.js
```diff
@@ -25,9 +25,17 @@ assert.ok(
     'Need to have GCLOUD_PROJECT defined ' +
         'along with valid application default credentials to be able to run this ' +
         'test');
+var config = {};
+// assert.ok(process.env.GCLOUD_TESTS_PROJECT_ID);
```
matthewloring: Commented code.

src/controller.js
cristiancavalli: Determinate variable declarations not at the top of the file? Although I would advocate for the consistent use of this in the same scope I'm not sure I agree with moving it out of the top-level of the function

src/index.js
cristiancavalli: We should probably also adopt a team-wide agreement on context aliasing. `self` or `that`. It might also be better to split across paradigms e.g. (OO & inheritance-based models use `self` while borrowed context references with outside invokers use `that`)

src/index.js
```diff
@@ -52,9 +51,10 @@ function Debug(options) {
   }
 
   var config = {
+    projectIdRequired: false,
     baseUrl: 'https://clouddebugger.googleapis.com/v2',
     scopes: [
-      // TODO: do we still need cloud-platform scope? 
+      // TODO: do we still need cloud-platform scope?
```
cristiancavalli: According to the API explorer cloud-debugger now has its own scope:
`https://www.googleapis.com/auth/cloud_debugger`

It would be interesting to see if this new scope fulfills all our needs

src/debuggee.js
ofrobots: Not really, but the code is going through a fair amount of refactoring so having tests would be beneficial at least in the short term. `projectId` and `uid` are required arguments. Checking for the type is a convenient way of ensuring that the required – until we get a bonafide type checking compiler in the pipeline.

src/index.js
ofrobots: Yep, that's why the TODO is there. I don't want to change this as part of this PR though.

src/index.js
ofrobots: Eh, this will come in a future refactor. I have removed this unused code for now.

ofrobots:
Ready for another look.

test/test-debuggee.js
```diff
@@ -0,0 +1,68 @@
...
+    assert.equal(debuggee.labels.module, 'some-service');
+    assert.equal(debuggee.labels.version, 'production');
+  });
+
+  it('should not add a module label when service is default', function() {
+    var debuggee = new Debuggee('fancy-project', 'very-unique',
+                                {service: 'default', version: 'yellow.5'});
+    assert.ok(debuggee);
+    assert.ok(debuggee.labels);
+    assert.equal(debuggee.labels.module, undefined);
```
cristiancavalli: This is technically a type comparison -- `assert.strictEqual`?

test/test-debuggee.js
```diff
@@ -0,0 +1,68 @@
...
+    assert.throws(function() { new Debuggee('test'); });
+    assert.throws(function() { new Debuggee('test', null); });
+  });
+
+  it('should have sensible labels', function() {
+    var debuggee = new Debuggee('some project', 'id', {
+      service: 'some-service',
+      version: 'production'
+    });
+    assert.ok(debuggee);
```
cristiancavalli: Since `assert.ok()` is just an alias of assert would it be clearer to just use `assert()`?

src/index.js
cristiancavalli: Gotcha

test/test-debuggee.js
cristiancavalli: sgtm, it does even the api surfaces

src/agent/debuglet.js
```diff
@@ -23,10 +23,10 @@ var util = require('util');
 var semver = require('semver');
 
 var v8debugapi = require('./v8debugapi.js');
-var DebugletApi = require('../debugletapi.js');
+var DebugletApi = require('../controller.js');
```
kjin: Is it necessary to change DebugletApi here to Controller?

kjin: (and anywhere else that Controller is referenced)

src/status-message.js
```diff
@@ -1,6 +1,6 @@
 /**
- * Copyright 2015 Google Inc. All Rights Reserved.
  *
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
kjin: Line move here doesn't resemble copyright notices elsewhere

src/controller.js
ofrobots: There is more refactoring coming here. I have chosen not to act on any of this just yet.

src/agent/debuglet.js
ofrobots: The variable name is being fixed in a follow-on change. The API class name was changed to match the official API description: https://cloud.google.com/debugger/api/reference/rest/.

PR #195, opened 2016-12-16
Refactor debuggee state out of controller and make Controller a ServiceObject

ofrobots:
* Try to run system and e2e tests on developer builds as well
* make Controller a ServiceObject
* Remove debuggee state from controller class

### DIFF
--- a/src/controller.js
+++ b/src/controller.js
@@ -21,7 +21,9 @@
  */
 
 var assert = require('assert');
+var common = require('@google-cloud/common');
 var qs = require('querystring');
+var util = require('util');
 
 /** @const {string} Cloud Debug API endpoint */
 var API = 'https://clouddebugger.googleapis.com/v2/controller';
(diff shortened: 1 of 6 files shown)


src/controller.js
```diff
@@ -30,32 +32,31 @@ var API = 'https://clouddebugger.googleapis.com/v2/controller';
  * @constructor
  */
 function Controller(debug) {
-  /** @priavate {Debug} */
-  this.debug_ = debug;
-
-  /** @private {string} debuggee id provided by the server once registered */
-  this.debuggeeId_ = null;
+  common.ServiceObject.call(this, {
+    parent: debug,
```
matthewloring: Does this need to be an instance of `ServiceObject`?

ofrobots: `parent` can be a `Service` or a `ServiceObject` as far as I understand.

PR #196, opened 2016-12-16
move config to src/agent and add jsdocs

ofrobots:


src/agent/config.js
```diff
@@ -0,0 +1,206 @@
...
+   */
+  description: undefined,
+
+  // FIXME(ofrobots): today we prioritize GAE_MODULE_NAME/GAE_MODULE_VERSION
+  // over the user specified config. We should reverse that.
+  /**
+   * @property {object=} Identifies the context of the running service -
+   * [ServiceContext](https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext?authuser=2).
+   * This information is utilized in the UI to identify all the running
+   * instances of your service. This is disocvered automatically when your
```
cristiancavalli: s/disocvered/discovered

src/agent/config.js
```diff
@@ -0,0 +1,206 @@
...
+   * application is running on Google Cloud Platform. You may optionally chose
```
cristiancavalli: t/chose/choose

src/agent/config.js
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * @property {number=} Number of seconds to wait after the
+     * `maxLogsPerSecond` rate is hit before logging resumes per logpoint.
+     * @default
+     */
+    logDelaySeconds: 1
+  },
+
+  //
```
cristiancavalli: Empty comment line

src/agent/config.js
```diff
@@ -0,0 +1,206 @@
...
+ */
+'use strict';
+
+/**
+ * @typedef {object} DebugAgentConfig
+ */
+
+module.exports = {
+  /**
+   * @property {boolean=} Whether the debug agent should be started.
```
matthewloring: I always understood the `=` syntax to indicate an optional parameter to a function. I don't think it makes sense for properties to be "optional". Maybe nullable `?` is better though I'm not a fan of a nullable boolean?

src/agent/config.js
```diff
@@ -0,0 +1,206 @@
...
+     * @default
+     */
+    logDelaySeconds: 1
+  },
+
+  // FIXME(ofrobots): stop accepting this property here
+  // A path to a key file relative to the current working directory. If this
+  // field is set, the contents of the pointed file will be used for
+  // authentication instead of your application default credentials.
+  keyFilename: null,
```
matthewloring: Some of these initialize with `null` and others with `undefined`. It would be good to go with one. My preference is `null`.

src/agent/config.js
ofrobots: Ack.

PR #199, opened 2016-12-17
Fill in unimplemented tests

matthewloring:


test/standalone/test-debuglet.js
```diff
@@ -145,7 +145,17 @@ describe(__filename, function(){
     debuglet.start();
   });
 
-  it('should error if a package.json doesn\'t exist');
+  it('should error if a package.json doesn\'t exist', function(done) {
+    var config = extend({}, defaultConfig, {projectId: 'fake-project', workingDirectory: __dirname});
+    debuglet = new Debuglet(fakeDebug, config, logger);
+
+    debuglet.once('initError', function(err) {
+      assert(err);
```
ofrobots: nit: prefer `assert.ifError(err)`  for readability.. The rest of the file has counter examples, but it would be good to prefer `assert` methods when possible.

test/standalone/test-debuglet.js
```diff
@@ -195,11 +205,88 @@ describe(__filename, function(){
     debuglet.start();
   });
 
-  it('should de-activate when the server responds with isDisabled');
+  it('should de-activate when the server responds with isDisabled', function(done) {
```
ofrobots: The test doesn't do precisely what this description specifies. It verifies that we are able re-register after a disabling by the server, but it doesn't verify that the debuglet went into a 'de-activate' state.

test/standalone/test-debuglet.js
```diff
@@ -195,11 +205,88 @@ describe(__filename, function(){
...
+    debuglet.once('registered', function(id) {
+      assert(id === DEBUGGEE_ID);
+      scope.done();
+      done();
+    });
+
+    debuglet.start();
+  });
+
+  it('should re-register when registration expires', function(done) {
```
ofrobots: The test doesn't seem to be doing what was advertised. This is testing whether we retry registration upon a 4xx error (specifically 404).

Registration is expired by a 404 response on the `breakpoints.list` API. See [here](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/c580792c1031f3fb51fdf9dc03ef1738db86f46c/src/agent/debuglet.js#L286).

test/standalone/test-debuglet.js
matthewloring: Done.

test/standalone/test-debuglet.js
matthewloring: I believe it does because if it didn't deactivate, it would attempt to fetch breakpoints which is an endpoint that nock does not provide causing the test to fail. This is subtle so I've added explicit `fetcherActive_` checks for clarity.

test/standalone/test-debuglet.js
matthewloring: Actually, sorry, I misread this. We want to fail if `err` is not an error. I do not see a corresponding `assert.error` call.

ofrobots: Actually I misread it. I think you want `assert.ok(err)`.

test/standalone/test-debuglet.js
```diff
@@ -195,11 +206,103 @@ describe(__filename, function(){
...
+      })
+      .post(REGISTER_PATH)
+      .reply(200, {
+        debuggee: {
+          id: DEBUGGEE_ID
+        }
+      });
+
+    debuglet.once('registered', function(id) {
+      assert(id === DEBUGGEE_ID);
```
ofrobots: nit: assert.equal

test/standalone/test-debuglet.js
```diff
@@ -195,11 +206,103 @@ describe(__filename, function(){
...
+      setImmediate(function() {
+        assert(debuglet.fetcherActive_);
```
ofrobots: nit: assert.ok

test/standalone/test-debuglet.js
```diff
@@ -195,11 +206,103 @@ describe(__filename, function(){
...
+        scope.done();
+        done();
+      });
+    });
+
+    setTimeout(function() {
+      assert(!debuglet.fetcherActive_);
```
ofrobots: ditto

matthewloring: Done.

test/standalone/test-debuglet.js
matthewloring: Done.

PR #200, opened 2016-12-19
credentials accepted in options only now

ofrobots:


src/index.js
```diff
@@ -63,12 +63,24 @@ function Debug(options) {
...
+  this.options = options;
 }
 util.inherits(Debug, common.Service);
 
 var initConfig = function(config_) {
   var config = config_ || {};
 
+  if (config.keyFilename || config.credentials || config.projectId) {
+    throw new Error('keyFilename, projectId or credentials should be provided' + 
+                    ' in the module options');
```
matthewloring: Do we use the term "module" options elsewhere in our documentation? This could be confusing if people aren't clear on the different options objects we deal with.

test/nocks.js
```diff
@@ -0,0 +1,53 @@
...
+  return nock('http://metadata.google.internal')
+    .get('/computeMetadata/v1/project/numeric-project-id')
+    .reply(reply);
+}
+
+module.exports = {
+  oauth2: nockOAuth2,
+  numericProjectId: nockNumericProjectId,
+  register: nockRegister
+};
```
matthewloring: `\n`

README.md
```diff
@@ -19,12 +19,25 @@ This module provides Stackdriver Debugger support for Node.js applications. [Sta
...
 # Install with `npm` or add to your `package.json`.
-npm install --save @google/cloud-debug
+npm install --save @google-cloud/debug
+```
 
-# Require and start the agent at the top of your main script (but after '@google/cloud-trace' if you are also using it).
-require('@google/cloud-debug').start();
+```js
+// Require and start in the startup of your application:
+var debug = require('@google-cloud/debug')();
```
matthewloring: Do we want to reference a package that does not yet exist?

README.md
```diff
@@ -123,8 +150,8 @@ As soon as that line of code is reached in any of the running instances of your
 [cloud-console-projects]: https://console.cloud.google.com/iam-admin/projects
 [app-default-credentials]: https://cloud.google.com/identity/protocols/application-default-credentials
 [service-account]: https://console.cloud.google.com/apis/credentials/serviceaccountkey
-[npm-image]: https://img.shields.io/npm/v/@google/cloud-debug.svg
-[npm-url]: https://npmjs.org/package/@google/cloud-debug
+[npm-image]: https://img.shields.io/npm/v/@google-cloud/debug.svg
```
matthewloring: I think we should update these all at once.

README.md
ofrobots: Okay, this is a valid concern because github would show the README from the default branch. There are two ways to solve this:

* change the default branch of the repo to (a newly created) 0.9.x branch. If we need to do a release between this module gets renamed we could also use this branch.
* Move all the recent refactoring to a 0.10.x branch and reset the master branch to be the latest 0.9.x

src/index.js
ofrobots: done.

test/nocks.js
ofrobots: done.

README.md
ofrobots: I went with the former option.

matthewloring:
LGTM once the unit tests are passing.

PR #201, opened 2016-12-20
Change e2e tests to use native Promises

kjin:
Removed dependency on `q`, replacing `q` promises with native promises.


kjin:
Q: Should I turn these into mocha tests?

matthewloring:
I understand that this will eventually become a system test in the google-cloud-node repo. Systems tests do seem to use mocha but it can probably be done in a follow on.

PR #207, opened 2016-12-22
Fix flakiness in test-debuglet.js

ofrobots:
* A setTimeout that doesn't dominate the async done() path causes flakiness.
* Improve the test to be more on-topic.

### DIFF
--- a/test/standalone/test-debuglet.js
+++ b/test/standalone/test-debuglet.js
@@ -223,6 +223,31 @@ describe(__filename, function() {
           id: DEBUGGEE_ID,
           isDisabled: true
         }
+      });
+
+    debuglet.once('remotelyDisabled', function() {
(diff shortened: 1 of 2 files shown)


test/standalone/test-debuglet.js
```diff
@@ -231,19 +256,19 @@ describe(__filename, function() {
...
       });
 
+    var gotDisabled = false;
+    debuglet.once('remotelyDisabled', function() {
+      assert.ok(!debuglet.fetcherActive_);
+      gotDisabled = true;
+    });
+
     debuglet.once('registered', function(id) {
+      assert.ok('gotDisabled');
```
matthewloring: Probably shouldn't be a string.

ofrobots: fixed.

PR #208, opened 2016-12-22
Add Debugger API to test/ and changed E2E tests to use them

kjin:
* `test/debugger.js` has an interface that makes calls to the Cloud Debugger API.
* `test/e2e/test-breakpoints.js` and `test/e2e/test-log-throttling.js` use this file.

### DIFF
--- a/test/debugger.js
+++ b/test/debugger.js
@@ -166,8 +166,8 @@ Debugger.prototype.getBreakpoint = function(debuggeeId, breakpointId, callback) 
     clientVersion: this.clientVersion_
   };
 
-  var uri = API + '/debuggees/' + encodeURIComponent(debuggee.id) +
-            '/breakpoints/' + encodeURIComponent(breakpoint.id) +
+  var uri = API + '/debuggees/' + encodeURIComponent(debuggeeId) +
+            '/breakpoints/' + encodeURIComponent(breakpointId) +
             '?' + qs.stringify(query);
   this.request({uri: uri, json: true}, function(err, body, response) {
     if (err) {
@@ -208,7 +208,7 @@ Debugger.prototype.setBreakpoint = function(debuggeeId, breakpoint, callback) {
     json: true,
     body: breakpoint
   };
-
+  
   this.request(options, function(err, body, response) {
     if (err) {
       callback(err);
--- a/test/debugger.js
+++ b/test/debugger.js
@@ -20,6 +20,7 @@
  * @module debug/debugger
  */
 
+var pjson = require('../package.json');
 var assert = require('assert');
 var common = require('@google-cloud/common');
 var qs = require('querystring');
@@ -41,7 +42,7 @@ function Debugger() {
   /** @private {string} */
   this.nextWaitToken_ = null;
 
-  this.clientVersion_ = 'google.com/cloud-debug/v1'; // TODO(kjin): Change this
+  this.clientVersion_ = pjson.name + '/default/v' + pjson.version;
 }
 
 util.inherits(Debugger, common.ServiceObject);
@@ -51,8 +52,8 @@ util.inherits(Debugger, common.ServiceObject);
  * breakpoints.
  * @param {string} projectId - The project ID for the project whose debuggees
  *     should be listed.
- * @param {bool} includeInactive - Whether or not to include inactive debuggees
- *     in the list.
+ * @param {boolean=} includeInactive - Whether or not to include inactive
+ *     debuggees in the list (default false).
  * @param {!function(?Error,Debuggee[]=)} callback - A function that will be
  *     called with a list of Debuggee objects as a parameter, or an Error object
  *     if an error occurred in obtaining it.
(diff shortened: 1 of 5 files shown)


test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+
+/** @const {string} Cloud Debug API endpoint */
+var API = 'https://clouddebugger.googleapis.com/v2/debugger';
+
+/**
+ * @constructor
+ */
+function Debugger() {
+  common.ServiceObject.call(this, {
+    parent: new Debug({}),
```
ofrobots: Should accept the Debug client as an parameter instead of creating from scratch.

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+    baseUrl: '/debugger'
+  });
+
+  /** @private {string} */
+  this.nextWaitToken_ = null;
+
+  this.clientVersion_ = pjson.name + '/default/v' + pjson.version;
```
ofrobots: In the controller I am changing `default` to `client`. Can you make the same change here?

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+      } else {
+        callback(null, []);
+      }
+    }
+  });
+};
+
+/**
+ * Gets a list of breakpoints in a given debuggee.
+ * @param {string} debuggeeId - The ID of the debuggee whose breakpoints should
```
ofrobots: question: Do we want to accept a `Debuggee` instead of the `debuggeeId`? Otherwise we should probably go change `Controller#listBreakpoints` and `Controller#updateBreakpoint` to accept the `debuggeeId` as well.

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+ *     be listed.
+ * @param {Object} options - An object containing options on the list of
```
ofrobots: s/Object/object/

Also, the options object is optional, so `{object=}`? (but I am not a jsdoc expert).

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+    query.waitToken = this.nextWaitToken_;
+  }
+
+  var uri = API + '/debuggees/' + encodeURIComponent(debuggeeId) +
+            '/breakpoints?' + qs.stringify(query);
+  this.request({uri: uri, json: true}, function(err, body, response) {
+    if (err) {
+      callback(err);
+    } else if (!response) {
+      callback(err || new Error('unknown error - request response missing'));
```
ofrobots: `err` guaranteed to be falsy here.

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+  };
+
+  var uri = API + '/debuggees/' + encodeURIComponent(debuggeeId) +
+            '/breakpoints/' + encodeURIComponent(breakpointId) +
+            '?' + qs.stringify(query);
+  this.request({uri: uri, json: true}, function(err, body, response) {
+    if (err) {
+      callback(err);
+    } else if (!response) {
+      callback(err || new Error('unknown error - request response missing'));
```
ofrobots: `err` guaranteed to be falsy here.

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+                         response.statusCode));
+    } else if (Object.keys(body).length > 0) {
+      callback(new Error('response body is non-empty'));
+    } else {
+      callback(null);
+    }
+  });
+};
+
+module.exports = Debugger;
```
ofrobots: Okay, this class is really well done, so I don't see a reason to keep it as a utility file for the tests. Please move this to `src/`.

test/debugger.js
```diff
@@ -0,0 +1,266 @@
...
+         qs.stringify(query),
+    method: 'DELETE',
+    json: true
+  };
+
+  this.request(options, function(err, body, response) {
+    if (err) {
+      callback(err);
+    } else if (!response) {
+      callback(err || new Error('unknown error - request response missing'));
```
ofrobots: `err` guaranteed to be falsy here.

test/e2e/package.json
```diff
@@ -10,6 +10,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "lodash": "^3.9.3",
-    "semver": "^5.3.0"
+    "semver": "^5.3.0",
+    "thenify-all": "^1.6.0"
```
ofrobots: @google-cloud/common has a promisifyAll function you could use and avoid this dependency.

https://github.com/GoogleCloudPlatform/google-cloud-node/blob/master/packages/common/src/util.js#L743

test/debugger.js
kjin: Personally I think it'd be better if we accepted both a debuggee object or simply its ID, and read the `id` property if it turns out to be an object (same goes for breakpoints). I went with just IDs first because `getBreakpoint` should take a breakpoint ID, and everything else followed suit. What do you think about making it accept either?

matthewloring: I think our api should match the service definition as closely as possible. I like the idea of having a single handle to a service object (if I get a debuggee back I should be able to pass it to any api call that needs a unique debuggee) but that should be a design choice made by the service, not our client library.

test/debugger.js
kjin: Great! I'll move this in a future PR.

test/debugger.js
kjin: Done

test/debugger.js
kjin: Done

test/debugger.js
kjin: Done

test/debugger.js
kjin: Done

test/debugger.js
kjin: Done

test/debugger.js
kjin: Done

test/debugger.js
kjin: In that case it should accept IDs for debuggees and breakpoints (like it does now), which means `Controller` will have to be updated as well.

ofrobots: If this is supposed to be a "Client Library" then this should be idiomatic to Node.js devs, as a general rule – so I don't agree that we should expose developers to service design choices.

I think the `debuggee` is the object that developers really care about – whether it contains an ID internally that the service uses to identify debuggees is not something that the users should care about.

But, I don't think this is the right time to be having a client library design discussion. I just want to make sure we have a consistent interface between the `Controller` and the `Debugger` APIs. Either both should accept ID or both should accept bona fide objects.

Speaking of the bona-fide objects, the list of debuggees returned isn't a list of `Debuggee` objects, it is just `Debuggee`-like (the controller api has similar issues). That would also be worth discussing when designing a proper client library interface.

PR #210, opened 2016-12-23
start using gcp-metadata for metadata queries

ofrobots:


src/agent/debuglet.js
```diff
@@ -263,30 +264,31 @@ Debuglet.createDebuggee =
...
-    Debuglet.prototype.getProjectId_ = function(callback) {
+/**
+ * @private
+ */
+Debuglet.prototype.getProjectId_ = function(callback) {
   var that = this;
 
   // We need to figure out whether we are running on GCP. We can use our ability
   // to access the metadata service as a test for that.
   // TODO: change this to getProjectId in the future.
```
matthewloring: Can this TODO go away now?

src/agent/debuglet.js
```diff
@@ -263,30 +264,31 @@ Debuglet.createDebuggee =
...
-                  metadataProject;
-
-    // We if don't have a projectId by now, we fail with an error.
-    if (!project) {
-      return callback(err);
-    }
-    return callback(null, project, onGCP);
-  });
+  metadata.project(
+      'numeric-project-id', function(err, response, metadataProject) {
```
matthewloring: Could we make this key available as a constant.

src/agent/debuglet.js
```diff
@@ -263,30 +264,31 @@ Debuglet.createDebuggee =
...
-    var project = that.debug_.options.projectId || process.env.GCLOUD_PROJECT ||
-                  metadataProject;
-
-    // We if don't have a projectId by now, we fail with an error.
-    if (!project) {
-      return callback(err);
-    }
-    return callback(null, project, onGCP);
-  });
+  metadata.project(
```
matthewloring: Does this have different retry behavior than what existed before?

test/standalone/test-debuglet.js
```diff
@@ -48,6 +48,7 @@ describe(__filename, function() {
   });
 
   it('should not start when projectId is not available', function(done) {
+    this.timeout(8000);
```
matthewloring: Yikes, is the new query approach slower?

src/agent/debuglet.js
ofrobots: In a follow-on change. Sure.

src/agent/debuglet.js
ofrobots: PRs on `gcp-metadata` are welcome!

test/standalone/test-debuglet.js
ofrobots: I think we should optionally disable retry in `gcp-metadata`, but I don't think it is required for this PR.

src/agent/debuglet.js
ofrobots: Yes.

src/agent/debuglet.js
matthewloring: My thought was to expose it as a constant within our own module but if it's only ever referenced here that's fine

PR #212, opened 2016-12-23
Use mocha for end-to-end tests

kjin:
Turned `test-breakpoints.js` and `test-log-throttling.js` into mocha tests in a new file `test-e2e.js`. This involved splitting off the debuggee into a separate file (`fib.js`).

Changes to the tests themselves are:
* Spawned children now each have their own log object in the test, and each log is individually checked (whereas before, there was a single object that contained logs from all children).
* The log throttling test now also spawns three children (instead of one), since it uses common initialization logic.


test/fixtures/fib.js
```diff
@@ -0,0 +1,54 @@
...
+// master.
+setTimeout(function() {
+  var ok = true;
+  function sendErrorIfNotOk(predicate, errorMessage) {
+    if (!predicate) {
+      ok = false;
+      process.send([errorMessage]);
+    }
+  };
+  sendErrorIfNotOk(debug.private_, 'debuglet has initialized');
```
matthewloring: If we fail on any of these it looks like we will send all following error messages. Could this be structured so that the process either receives a single error or valid debuggee information.

test/e2e/test-e2e.js
```diff
@@ -0,0 +1,323 @@
...
+
+var FILENAME = 'test/fixtures/fib.js';
+
+var delay = function(delayTimeMS) {
+  return new Promise(function(resolve, reject) {
+    setTimeout(resolve, delayTimeMS);
+  });
+};
+
+describe('@google-cloud/debug end-to-end behavior (allow 60s)', function () {
```
matthewloring: Maybe 'allow 60s' as a comment instead of part of test name?

test/fixtures/fib.js
```diff
@@ -0,0 +1,54 @@
...
+  var debuglet = debug.private_;
+  var debuggee = debuglet.debuggee_;
+  sendErrorIfNotOk(debuggee, 'should create debuggee');
+  sendErrorIfNotOk(debuggee.project, 'debuggee should have a project');
+  sendErrorIfNotOk(debuggee.id, 'debuggee should have registered');
+  if (ok) {
+    // The parent process needs to know the debuggeeId and project.
+    process.send(['', debuggee.id, debuggee.project]);
```
matthewloring: Could this be changed to send an object which optionally defines the properties `error`, `debuggeeId`, and `debuggeeProject`? That makes the case checks easier to read (more than `a[0].length > 0`).

test/e2e/test-e2e.js
```diff
@@ -0,0 +1,323 @@
...
+        if (!debuggeeId) {
+          // Cache the needed info from the first worker.
+          debuggeeId = a[1];
+          projectId = a[2];
+        } else {
+          // Make sure all other workers are consistent.
+          assert.equal(debuggeeId, a[1]);
+          assert.equal(projectId, a[2]);
+          if (debuggeeId !== a[1] || projectId !== a[2]) {
+            reject(new Error('Child debuggee ID and/or project ID' +
```
matthewloring: It might be good to add a return after the `reject` here and above since the function continues executing even after the call (it happens to not matter here because even if it does run the resolve the promise will have already rejected).

test/e2e/test-e2e.js
```diff
@@ -0,0 +1,323 @@
...
+        // Each child has its own transcript.
+        child.process.stdout.on('data', stdoutHandler(i));
+        child.process.stderr.on('data', stdoutHandler(i));
+      }
+    });
+  });
+
+  afterEach(function() {
+    children.forEach(function (child) {
+      child.process.kill();
```
matthewloring: We may also want to wait for the child process to exit here. It may be possible for the final error messages printed by the children of one test to show up on the next test's transcripts.

kjin: Done.

test/e2e/test-e2e.js
kjin: Done

test/e2e/test-e2e.js
kjin: Done

test/fixtures/fib.js
kjin: Done

test/fixtures/fib.js
kjin: Oops! Done

test/e2e/test-e2e.js
kjin: I think it's technically not possible for children to print error messages after the beforeEach hook is resolved, but this will reduce the possibility of tests interfering with one another.

kjin:
Still fixing tests - since E2E tests fail.

kjin:
E2E tests pass: https://travis-ci.org/GoogleCloudPlatform/cloud-debug-nodejs/builds/189010678

test/standalone/test-capture-time.js
```diff
@@ -0,0 +1,109 @@
+// /**
```
ofrobots: Did you intend to move this and then comment it out?

kjin: Oops, not sure why I commented this out

kjin: Uncommented

kjin: So now that I have uncommented and run the tests I realized why I commented it out in the first place - `test-capture-time.js` frequently failed (as far as I have seen) because it has fairly strict requirements on the capture time (<20ms) that can't be met. I was wondering, should I relax this requirement?

ofrobots: I don't think this should be considered a test at all. I think it is more of a performance benchmark that needs a bit more polish before it is instated as a requirement. I think this should be removed / moved out of the way for now.

PR #213, opened 2016-12-27
fix stale code/test from test-controller

ofrobots:
The controller no longer automatically fetches the projectId from metadata service. The mock and corresponding tests were no longer necessary.


ofrobots:
Yep. `test-debuglet.js` exercises this path adequately.

PR #214, opened 2017-01-06
Move some tests out of test/standalone

kjin:
Moved these tests out of `test/standalone` (with path changes and minor fixes to make them run independently of each other):

- `test-debuglet.js`
- `test-module.js`
- `test-options-credentials.js`
- `test-sourcemapper.js`

Also, moved `test-v8debugapi.js` back to `test/standalone`. Because it's always run last it didn't seem to have interference with other tests. But now that `test-module.js` is non-interference, `test-v8debugapi.js` fails unless being run alone.

All of the remaining standalone tests have to do with the `v8debugapi` object, which is a singleton. To make these tests standalone would involve adding a debug-only function to `src/v8debugapi.js` to modify the singleton... but was not sure if this would be a reasonable thing to do. Opinions wanted here.


ofrobots:
I'd be okay with adding a test-only API into `src/v8debugapi.js` to make the tests stateless. We don't export this file outside the module, so I don't have concerns about the misuse of the test-only API.

kjin:
Sounds good, I will address that in another PR.

kjin:
@ofrobots As it turns out, having tests being non-interfering also requires a change in `index.js` itself, to override the `debuglet` singleton (see [here](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/src/index.js#L86)). Do you think I should go ahead with this, and if so, what would be a more suitable approach: exposing a test-only `stop` method, or enabling a test-only `force` flag that causes `debuglet` to be re-assigned when set?

ofrobots:
I like the idea of a test-only `stopAgent` method better, but I don't have really strong feelings on this.

PR #215, opened 2017-01-07
Removed dependency on @google/cloud-diagnostics-common

kjin:
Removed all references to `@google/cloud-diagnostics-common`, relying on equivalent functionality in `@google-cloud/common`.

To facilitate this a new augmented `debug-logger` class is introduced. I'm proposing to handle breakpoint and interval logging slightly differently than before, PTAL.

1/9: Also removed references to metadata endpoint `numeric-project-id`, and added tests to check that `GCLOUD_PROJECT` and `config.projectId` are both respected. May be helpful to look at the two commits separately.


test/nocks.js
```diff
@@ -40,14 +40,14 @@ function nockRegister(validator) {
       .reply(200);
 }
 
-function nockNumericProjectId(reply) {
+function nockProjectId(reply) {
   return nock('http://metadata.google.internal')
-    .get('/computeMetadata/v1/project/numeric-project-id')
+    .get('/computeMetadata/v1/project/project-id')
     .reply(reply);
```
cristiancavalli: Add `.once` or .`.times(x)` can be helpful when looking at a nock instance 4-6 months down the line and try to understand its longevity.

test/test-debuglet.js
```diff
@@ -112,6 +113,7 @@ describe('Debuglet', function() {
 
       debuglet.once('registered', function(id) {
         assert.equal(id, DEBUGGEE_ID);
+        assert.equal(debuglet.debuggee_.project, projectId);
```
cristiancavalli: For value comparisons I normally advocate for the use of `assert.strictEqual` to avoid 'truthy' equivocations but this is totally preference.

test/test-debuglet.js
```diff
@@ -120,6 +122,61 @@ describe('Debuglet', function() {
       debuglet.start();
     });
 
+    it('should use GCLOUD_PROJECT in lieu of config.projectId', function(done) {
+      process.env.GCLOUD_PROJECT = '11020304f2934-b';
```
cristiancavalli: When setting this do you save the old value and then restore it? I can't seem to find it in this change. Only reason I advocate for this is this was a huge pain in growing errors tests where multiple test files ended up using and modifying the `env`

test/test-debuglet.js
kjin: Agreed. Currently the vast majority of comparisons are still `assert.equal`, and I added similar comparisons for uniformity; I think it would be best to change it all at once (in another PR). I don't think it's harmful as of right now since the RHS is almost always a string constant.

src/debug-logger.js
```diff
@@ -0,0 +1,71 @@
...
+  var newLogger = common.logger(options);
+
+  common.logger.LEVELS.forEach(function(logLevel) {
+    newLogger[logLevel].breakpoint = function(msg, breakpoint) {
+      return newLogger[logLevel](formatBreakpoint(msg, breakpoint));
+    };
+    newLogger[logLevel].interval = function(msg, interval) {
+      return newLogger[logLevel](formatInterval(msg, interval));
+    };
+  });
```
ofrobots: This looks a bit heavy handed for a fairly simple problem. I think you can move `formatBreakpoint` function to `debuglet.js` (it is only needed there), and use `logger_.info` on the resulting string. Same for `interval` and `v8debugapi.js`.

Secondly, it seems like you have a type mismatch. `logger_.info.breakpoint` is called on breakpoint map whereas the function seems to be expecting a single breakpoint. Am I reading this wrong?

test/test-debuglet.js
ofrobots: +1. But instead of saving restoring, simply add a `projectId` property to the config object below.

ofrobots: my suggestion is invalid now that I read the title of the test :)

kjin: The test happens to already `delete process.env.GCLOUD_PROJECT` before each test case. But I can add saving/restoring logic there.

src/debug-logger.js
kjin: I realized that the original logger had both `breakpoint` and `breakpoints`, and I implemented `breakpoint` but only `breakpoints` is used. So yes, there's a type mismatch.

Will address that and also move `format*`.

kjin:
@ofrobots PTAL

src/debug-logger.js
```diff
@@ -0,0 +1,71 @@
+/**
```
ofrobots: Is this file still needed?

system-test/test-e2e.js
```diff
@@ -24,7 +24,7 @@ var promisifyAll = require('@google-cloud/common').util.promisifyAll;
 var Debug = require('..');
 var Debugger = require('../test/debugger.js');
 
-var CLUSTER_WORKERS = 3;
+var CLUSTER_WORKERS = 1;
```
ofrobots: Did you accidentally check this in?

kjin: Yes, will revert

src/debug-logger.js
kjin: No, thanks for the catch!

kjin:
@ofrobots Sounds good, I'll try to keep future PRs more singular in purpose.

PR #217, opened 2017-01-13
Test success_on_timeout behavior

matthewloring:


matthewloring:
It turns out the service accepts either `success_on_timeout` or `successOnTimeout` as the query parameter. I've added a test ensuring that things work as intended. PTAL.

PR #218, opened 2017-01-13
Make all tests runnable together

kjin:
Moved remaining `standalone tests` to the root `test` folder.

The simplest way to go about this was to introduce an undocumented, debug-only flag `force_`, that allows both the debug agent and V8 debug API singletons to be re-assigned. I'm open to doing this a different way, however, like a `stop` function.

### DIFF
--- a/src/agent/v8debugapi.js
+++ b/src/agent/v8debugapi.js
@@ -59,7 +59,7 @@ var formatInterval = function(msg, interval) {
 
 var singleton;
 module.exports.create = function(logger_, config_, jsFiles_, sourcemapper_) {
-  if (singleton) {
+  if (singleton && !config_.force) {
     return singleton;
   }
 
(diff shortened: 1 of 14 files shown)


src/agent/v8debugapi.js
```diff
@@ -59,7 +59,7 @@ var formatInterval = function(msg, interval) {
 
 var singleton;
 module.exports.create = function(logger_, config_, jsFiles_, sourcemapper_) {
-  if (singleton) {
+  if (singleton && !config_.force_) {
```
matthewloring: Could we change the name to be more explicit even if this is for testing only? Maybe `forceNewAgent_`?

kjin: Yes, that sounds better - done.

kjin:
@ofrobots @matthewloring PTAL

PR #219, opened 2017-01-17
Update README markdown

cristiancavalli:
Fix formatting so markdown headers for languages don't render.


cristiancavalli:

PR #220, opened 2017-01-17
Remove misleading arguments/locals message

cristiancavalli:
When within the frame capture range omit status
messages declaming that arguments/locals are only
captured for frames in the capture range.

### DIFF
--- a/src/agent/state.js
+++ b/src/agent/state.js
@@ -224,11 +224,12 @@ StateResolver.prototype.resolveFrames_ = function() {
   var frames = [];
   var frameCount = Math.min(this.state_.frameCount(),
     this.config_.capture.maxFrames);
+    
   for (var i = 0; i < frameCount; i++) {
     var frame = this.state_.frame(i);
     if (this.shouldFrameBeResolved_(frame)) {
-      var resolveVars = i < this.config_.capture.maxExpandFrames;
-      frames.push(this.resolveFrame_(frame, resolveVars));
+      frames.push(this.resolveFrame_(frame,
+        (i < this.config_.capture.maxExpandFrames)));
     }
   }
   return frames;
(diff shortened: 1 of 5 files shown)


cristiancavalli:

src/agent/state.js
```diff
@@ -286,15 +288,16 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
   return path.indexOf('node_modules') === 0;
 };
 
-StateResolver.prototype.resolveFrame_ = function(frame, resolveVars) {
-  var noArgs = [{
+StateResolver.prototype.resolveFrame_ = function(frame, resolveVars, currentFrame, totalFrames) {
+  var framesExceeded = (currentFrame >= totalFrames);
```
ofrobots: Looking at all the code-paths through here, I believe `framesExceeded` will always be equal to `!resolveVars`. Am I reading things correctly?

cristiancavalli:
@ofrobots thanks for the catch. `resolveFrames_` has been updated to utilize this elision.

src/agent/state.js
```diff
@@ -286,32 +287,44 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
...
-    args = resolveVars && (!isEmpty(args)) ? this.resolveArgumentList_(args) :
-      noArgs;
+  };
+};
+
+StateResolver.prototype.resolveFrame_ = function(frame, underFrameCap) {
+  var args = [];
+  var locals = [];
+  if (!underFrameCap) {
+    args.push(this.produceArgumentsNotAvailableMessage());
```
ofrobots: Since there is no other use of a constant object building method `produceArgumentsNotAvailable`, just inline the object here and get rid of the function.

src/agent/state.js
```diff
@@ -286,32 +287,44 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
...
+    locals.push(this.produceLocalsNotAvailableMessage());
```
ofrobots: ditto.

cristiancavalli:
updated

PR #221, opened 2017-01-19
Support for AppEngine Flex environment variables

ofrobots:
Also try to detect the serviceContext from the environment variables.

Apologies for the style changes. Accidentally used the 'global format' short-cut in the editor. I can try and remove the style changes if necessary.

### DIFF
--- a/test/test-debuglet.js
+++ b/test/test-debuglet.js
@@ -36,32 +36,26 @@ var oldGP;
 var bp = {
   id: 'test',
   action: 'CAPTURE',
-  location: { path: 'fixtures/foo.js', line: 2 }
+  location: {path: 'fixtures/foo.js', line: 2}
 };
 var errorBp = {
   id: 'testLog',
   action: 'FOO',
-  location: { path: 'fixtures/foo.js', line: 2 }
+  location: {path: 'fixtures/foo.js', line: 2}
 };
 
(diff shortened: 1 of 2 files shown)


test/test-debuglet.js
```diff
@@ -553,26 +528,21 @@ describe('Debuglet', function() {
...
+              .reply(200, {breakpoints: [bp]})
+              .put(BPS_PATH + '/test',
+                   function(body) {
+                     return body.breakpoint.status.description.format ===
+                            'The snapshot has expired';
+                   })
+              .reply(200)
+              .get(BPS_PATH + '?success_on_timeout=true')
+              .times(4)
+              .reply(200, {breakpoints: [bp]});
```
kjin: Were these whitespace changes intended? The way they occur in code that was otherwise not changed seems to suggest that they were auto-formatted by the editor.

ofrobots: These were autoformatted by the editor. I accidentally used to 'global format' rather than 'changed lines'. :(

ofrobots:
yarn.lock dropped.

matthewloring:
LGTM once the tests are green

ofrobots:
The 0.12 build needed to be restarted. Green now.

PR #223, opened 2017-01-20
Fix bugs in findScripts

ofrobots:
* Code was buggy when `fileStats` had only one entry.
* Fix bug in `pathToRegExp` that was not considering '.' as literal.
* Make the code easier to unit test.
* More unit tests.

### DIFF
--- a/src/agent/v8debugapi.js
+++ b/src/agent/v8debugapi.js
@@ -58,7 +58,7 @@ var formatInterval = function(msg, interval) {
 };
 
 var singleton;
-module.exports.create = function(logger_, config_, jsFiles_, sourcemapper_) {
+function create(logger_, config_, jsFiles_, sourcemapper_) {
   if (singleton && !config_.forceNewAgent_) {
     return singleton;
   }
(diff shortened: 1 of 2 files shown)


ofrobots:
PTAL.

src/agent/v8debugapi.js
```diff
@@ -585,4 +564,67 @@ module.exports.create = function(logger_, config_, jsFiles_, sourcemapper_) {
...
+  scriptPath = scriptPath.replace('.', '\\.');
+  if (path.sep !== '/') {
+    scriptPath = scriptPath.replace(new RegExp('\\\\', 'g'), '\\\\');
+  }
+  return new RegExp(scriptPath + '$');
+}
+
+/**
+ * Given an list available files and a script path to match, this function tries
+ * to resolve the script patch a (hopefully unique) match in the file list
```
matthewloring: `s/patch/path`

src/agent/v8debugapi.js
```diff
@@ -585,4 +564,67 @@ module.exports.create = function(logger_, config_, jsFiles_, sourcemapper_) {
...
+ * irrespective of the script path. This can be useful because repo file paths
```
matthewloring: Not sure what you mean by `irrespective of the script path` here.

src/agent/v8debugapi.js
ofrobots: done.

src/agent/v8debugapi.js
ofrobots: done.

PR #224, opened 2017-01-20
Describe enabling Debugging API access scopes for GCE instances in README

kjin:
The GCP console doesn't automatically enable the Stackdriver Debugger API for new GCE instances with the default options, so I've made changes to the README describing how to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ the [Google Cloud Console][dev-console]. You can start adding snapshots and log-
 
 ## Running on Google Cloud Platform
 
-The Stackdriver Debugger Agent should work out of the box for Google App Engine instances, as long as the [Stackdriver Debugger API][debugger-api] is enabled on your project (this is the default).
+The Stackdriver Debugger Agent should work without manually provided authentication credentials for instances running on Google Cloud Platform, as long as the [Stackdriver Debugger API][debugger-api] access scope is enabled on that instance. For Google App Engine instances, this is automatic if the Debugger API has been enabled for your project (which it is by default).
 
-For new Google Compute Engine instances, you need to create the instance with the Stackdriver Debugger API access scope. From the web UI, you can do this in one of two ways under **Identity and API access**:
+For new Google Compute Engine instances, you need to explicitly enable the Debugger API access scope. From the web UI, you can do this in one of two ways under **Identity and API access**:
 * Use the Compute Engine default service account "Allow full access to all Cloud APIs" under Access scopes.
 * Select a service account with the **Cloud Debugger Agent** role (or any role with at least the same permissions). You may need to create one if you don't have one already.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,7 +41,13 @@ the [Google Cloud Console][dev-console]. You can start adding snapshots and log-
 
 ## Running on Google Cloud Platform
 
-Stackdriver debugger agent should work out of the box if your code is running on Google Cloud Platform, just make sure that the [Stackdriver Debugger API][debugger-api] is enabled on your project (this is the default).
+The Stackdriver Debugger Agent should work without manually provided authentication credentials for instances running on Google Cloud Platform, as long as the [Stackdriver Debugger API][debugger-api] access scope is enabled on that instance. For Google App Engine instances, this is automatic if the Debugger API has been enabled for your project (which it is by default).
+
+For new Google Compute Engine instances, you need to explicitly enable the Debugger API access scope. From the web UI, you can do this in one of two ways under **Identity and API access**:
+* Use the Compute Engine default service account and select "Allow full access to all Cloud APIs" under Access scopes.
+* Select a service account with the **Cloud Debugger Agent** role (or any role with at least the same permissions). You may need to create one if you don't have one already.
```
matthewloring: What is meant by "or any role with at least the same permissions"?

README.md
```diff
@@ -41,7 +41,13 @@ the [Google Cloud Console][dev-console]. You can start adding snapshots and log-
 
 ## Running on Google Cloud Platform
 
-Stackdriver debugger agent should work out of the box if your code is running on Google Cloud Platform, just make sure that the [Stackdriver Debugger API][debugger-api] is enabled on your project (this is the default).
+The Stackdriver Debugger Agent should work without manually provided authentication credentials for instances running on Google Cloud Platform, as long as the [Stackdriver Debugger API][debugger-api] access scope is enabled on that instance. For Google App Engine instances, this is automatic if the Debugger API has been enabled for your project (which it is by default).
+
+For new Google Compute Engine instances, you need to explicitly enable the Debugger API access scope. From the web UI, you can do this in one of two ways under **Identity and API access**:
```
matthewloring: Maybe we could link to the API access page here.

README.md
kjin: At least the same permissions that are granted for the Cloud Debugger Agent role. I didn't want to repeat the name of the role in the same sentence, but since it's unclear I'll expand on it.

README.md
kjin: It's a subsection of the create new GCE page. I could link to that page or include a screenshot - what do you think? I'm leaning towards the latter, it seems strange to add a link to the page used to create new instances.

matthewloring: Ah, I assumed "web UI" to refer to the console as a whole. Maybe specifying the GCE creation page would be sufficient as well.

PR #227, opened 2017-01-23
test: reduce dependence on repo layout

ofrobots:


test/test-debuglet.js
```diff
@@ -309,10 +308,14 @@ describe('Debuglet', function() {
...
 
     it('should pass source context to api if present', function(done) {
-      process.chdir(path.join(__dirname, 'fixtures'));
-
       var debug = require('..')(
           {projectId: 'fake-project', credentials: fakeCredentials});
+      var old = Debuglet.prototype.getSourceContext_;
+      Debuglet.prototype.getSourceContext_ = function(cb) {
+        setImmediate(function () {
+          cb(null, {a: 5});
```
matthewloring: Do we want to delete the source context in fixtures as well if it is no longer used?

ofrobots: Yep. I meant to do that but missed. Thanks for catching.

PR #237, opened 2017-02-07
provide minorversion label on AppEngine

ofrobots:
AppEngine flex no longer provides the GAE_MINOR_VERSION environment variable, but the debug UI depends upon it. Provide a fake value, but only on flex.

### DIFF
--- a/src/agent/debuglet.js
+++ b/src/agent/debuglet.js
@@ -144,7 +144,14 @@ Debuglet.prototype.normalizeConfig_ = function(config) {
     logLevel: process.env.GCLOUD_DEBUG_LOGLEVEL,
     serviceContext: {
       service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
-      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION
+      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+      // Debug UI expects GAE_MINOR_VERSION to be available for AppEngine, but
+      // AppEngine Flex doesn't have this environment variable. We provide a
+      // fake value as a work-around, but only on Flex (GAE_SERVICE will be
+      // defined on Flex).
+      minorVersion:
+          process.env.GAE_MINOR_VERSION ||
+              (process.env.GAE_SERVICE ? 'fake-minor-version' : undefined)
     }
   };
 
@@ -272,6 +279,11 @@ Debuglet.createDebuggee =
     if (_.isString(serviceContext.version)) {
       labels.version = serviceContext.version;
       desc += ' version:' + serviceContext.version;
+    }
+
+    if (_.isString(serviceContext.minorVersion)) {
+      //          v--- intentional lowercase
+      labels.minorversion = serviceContext.minorVersion;
     }
   }
 
--- a/test/test-debuglet.js
+++ b/test/test-debuglet.js
@@ -15,6 +15,7 @@
  */
 'use strict';
 
+var _ = require('lodash');
 var assert = require('assert');
 var defaultConfig = require('../src/agent/config.js');
 var Debuglet = require('../src/agent/debuglet.js');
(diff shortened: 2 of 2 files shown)


test/test-debuglet.js
```diff
@@ -243,6 +244,78 @@ describe('Debuglet', function() {
...
+           assert.ok(_.isString(debuglet.config_.serviceContext.minorVersion));
+         });
+
+      it('should not have minorVersion unless enviroment provides it',
+         function() {
+           var debug = require('../src/debug.js')();
+           var debuglet = new Debuglet(debug, defaultConfig);
+           assert.ok(debuglet.config_);
+           assert.ok(debuglet.config_.serviceContext);
+           assert.ok(!debuglet.config_.serviceContext.minorVersion);
```
cristiancavalli: would `assert.fail` be more idiomatic here in context to the assert lib?

test/test-debuglet.js
```diff
@@ -243,6 +244,78 @@ describe('Debuglet', function() {
...
+             debuglet.stop();
+             scope.done();
+             done()
+           });
+           debuglet.start();
+         });
+
+      it('should provide minorversion upon registration if on flex', function(
+                                                                         done) {
+        process.env.GAE_SERVICE = 'fake-service';
```
cristiancavalli: Should these env sets be rolled back after this test module exits?

test/test-debuglet.js
```diff
@@ -243,6 +244,78 @@ describe('Debuglet', function() {
...
+           assert.ok(!debuglet.config_.serviceContext.minorVersion);
+         });
+
+      it('should not provide minorversion upon registration on non flex',
+         function(done) {
+           var debug = require('../src/debug.js')(
+               {projectId: 'fake-project', credentials: fakeCredentials});
+           var debuglet = new Debuglet(debug, defaultConfig);
+
+           var scope = nock(API).post(REGISTER_PATH, function(body) {
```
cristiancavalli: Adding a `.once()` (or `twice` or `thrice` -- whatever it is) here to the scope may help with trying to figure out how long the scope is supposed to stay alive when looking at this 6 months down the road.

test/test-debuglet.js
cristiancavalli: or an strict assertion of value?

test/test-debuglet.js
ofrobots: See the `beforeEach` and `afterEach` section for this `describe`.

test/test-debuglet.js
ofrobots: Doesn't `assert.fail` always throw? That's not the behaviour I want here. `assert.ifError` is more appropriate behaviour, but less readable.

src/agent/debuglet.js
```diff
@@ -273,6 +280,11 @@ Debuglet.createDebuggee =
       labels.version = serviceContext.version;
       desc += ' version:' + serviceContext.version;
     }
+
+    if (_.isString(serviceContext.minorVersion)) {
+      //          v--- intentional lowercase
```
matthewloring: Is this documented anywhere?

test/test-debuglet.js
cristiancavalli: @ofrobots You're right; my bad. I agree that it's less readable.

test/test-debuglet.js
ofrobots: Done

test/test-debuglet.js
ofrobots: Wrote it slightly differently that might be a bit better.

test/test-debuglet.js
```diff
@@ -243,6 +244,78 @@ describe('Debuglet', function() {
...
+           var debuglet = new Debuglet(debug, defaultConfig);
+           assert.ok(debuglet.config_);
+           assert.ok(debuglet.config_.serviceContext);
+           assert.strictEqual(debuglet.config_.serviceContext.minorVersion,
+                              'some minor version');
+         });
+
+      it('should conjure a fake minor version when running on flex',
+         function() {
+           process.env.GAE_SERVICE = 'fake-gae-service';
```
DominicKramer: Perhaps we should have tests that verify that if `GAE_MINOR_VERSION` is either the empty string or `undefined`, then the `GAE_SERVICE` value is analyzed.

ofrobots: This test does verify the `undefined` value case (`process.env` is cleared before the test). I do not believe we should provide a fake minor version when the environment variable has a valid value albeit an empty string.

ofrobots:
The fix seems to be working.

PR #240, opened 2017-02-10
merge configs using a deep copy

ofrobots:
We were losing the default values of some properties previously.

This depends upon #239. That commit will also show up here until that lands.


ofrobots:
TBR= @matthewloring

matthewloring:
LGTM.

PR #244, opened 2017-02-22
Add allowExpressions option

matthewloring:


test/test-v8debugapi.js
```diff
@@ -915,6 +916,30 @@ describe('v8debugapi', function() {
         });
     });
 
+    it('should capture without values when `allowExpressions=false`', function(done) {
+      // clone a clean breakpointInFoo
+      var bp = {id: breakpointInFoo.id, location: breakpointInFoo.location};
+      config.allowExpressions = false;
```
ofrobots: better: `var config = extend(config, { allowExpressions: false })`

test/test-debuglet.js
```diff
@@ -526,6 +527,88 @@ describe('Debuglet', function() {
       debuglet.start();
     });
 
+    it('should reject breakpoints with conditions when allowExpressions=false',
```
ofrobots: nit: It would be better+simpler to add this test to `test-v8debug.js`.

test/test-debuglet.js
```diff
@@ -526,6 +527,88 @@ describe('Debuglet', function() {
...
+          debuglet.config_.allowExpressions = true;
+          scope.done();
+          done();
+        }, 1000);
+      });
+
+      debuglet.start();
+    });
+
+    it('should reject breakpoints with expressions when allowExpressions=false',
```
ofrobots: ditto.

src/agent/state.js
```diff
@@ -290,7 +296,16 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
...
   var args = [];
   var locals = [];
-  if (!underFrameCap) {
+  if (!this.config_.allowExpressions) {
+    args.push({
+      name: 'arguments_not_available',
+      varTableIndex: EXPRESSIONS_DISALLOWED_INDEX
+    });
+    locals.push({
+      name: 'locals_not_available',
```
ofrobots: Remind me again why looking at local values can have side effects? I do agree that we shouldn't look at object properties.

test/test-v8debugapi.js
matthewloring: That creates a new `config` object. This works by updating the values on the global config which is held by reference in the `api`. If we use `extend` we will need comparable lines replacing/restoring the config object held by the api.

test/test-debuglet.js
matthewloring: I can't find a `test-v8debug.js`, do you mean `test-v8debugapi.js`? The issue is that the logic being testing is in `debuglet.js` so I'm not sure it is testable from `test-v8debugapi.js`.

test/test-debuglet.js
matthewloring: same as above.

src/agent/state.js
matthewloring: I remember talking through it when we talked about the behavior of the new option being added but I don't recall the reason. We already avoid interceptors/getters on objects by default so object properties should be ok if locals are in general?

matthewloring: After digging back through some of the frame mirror resolution logic, I think locals/args are fine. I've added a comment justifying this.

src/agent/config.js
```diff
@@ -37,6 +37,15 @@ module.exports = {
    */
   description: null,
 
+  /**
+   * @property {boolean} Whether or not it is safe to evaluate expressions.
```
ofrobots: nit: s/safe/permitted

src/agent/state.js
```diff
@@ -290,6 +290,8 @@ StateResolver.prototype.isPathInNodeModulesDirectory_ = function(path) {
 StateResolver.prototype.resolveFrame_ = function(frame, underFrameCap) {
   var args = [];
   var locals = [];
+  // Locals and arguments are safe to collect even when `config.allowExpressions=false`
+  // since we properly avoid inspecting interceptors and getters by default.
```
ofrobots: Are there tests with a getter, setter, interceptors that will keep us sane?

matthewloring: Yup: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/48a9952ec004e18acb9e5ca876a411493b742129/test/test-v8debugapi.js#L657

src/agent/config.js
matthewloring: done

PR #245, opened 2017-02-24
Relax configuration rules for agent

matthewloring:
With this change, we will use the option object for agent configuration if options.debug does not exist.


matthewloring:
Done.

Issue #263, opened 2017-05-17
Failed to re-register debuggee XXX: ApiError: Precondition check failed.


stoianchoo:
GAE Flex NodeJS

`"node": ">6.4.0"`

`require('@google-cloud/debug-agent').start({ allowExpressions: true });`
after the trace-agent requre.

I get an error:
ERROR:@google-cloud/debug-agent: Failed to re-register debuggee project-name-redacted: ApiError: Precondition check failed.

_mod-edit: redacted project name._

stoianchoo:
The error is gone now. I think there was an update to the debugger and now I see the "logpoint" functionality that I don't remember seeing before.

Still the error message did not help in understanding the problem.

ofrobots:
@stoianchoo Thanks for the report & feedback. We're looking at how to improve the error message.

I am very curious about the error. A few questions:
* Did this occur at application startup, or sometime later?
* What version of the debug-agent are you using?

stoianchoo:
@ofrobots Just in case you still need this info:
- Not sure exactly when it occurred but it was definitely occurring when the instance was restarted.
- "@google-cloud/debug-agent": "^1.0.0"

ofrobots:
Thanks! I believe this should be fixed by #265. We'll try to get that released into a new version of the module in the coming week.

PR #265, opened 2017-05-18
stringify sourceContext before concat

ofrobots:
When sourceContext changed, our uniquifier wasn't because of a bug in how we were concatenating it.

Fixes: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/issues/263

### DIFF
--- a/src/agent/debuglet.js
+++ b/src/agent/debuglet.js
@@ -290,9 +290,8 @@ Debuglet.createDebuggee =
     desc += ' description:' + description;
   }
 
-  var uniquifier = desc + version + uid + JSON.stringify(sourceContext) + 
-      JSON.stringify(labels);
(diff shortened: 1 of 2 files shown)


ofrobots:
Yep - already added.

src/agent/debuglet.js
```diff
@@ -736,3 +735,11 @@ Debuglet._delimit = function(source, delim) {
   }
   return dest;
 };
+
+Debuglet._createUniquifier = function (desc, version, uid, sourceContext,
+  labels) {
+  var uniquifier = desc + version + uid + JSON.stringify(sourceContext) +
+    JSON.stringify(labels);
+  uniquifier = crypto.createHash('sha1').update(uniquifier).digest('hex');
```
matthewloring: Nit: could we return `crypto.createHash....` directly to avoid reassigning just to return?

PR #269, opened 2017-06-05
Update truncated object message

matthewloring:
The message has been updated to indicate that using the object in an
expression will bypass truncation limits.

### DIFF
--- a/src/agent/state.js
+++ b/src/agent/state.js
@@ -36,7 +36,6 @@ var BUFFER_FULL_MESSAGE_INDEX = 0;
 var NATIVE_PROPERTY_MESSAGE_INDEX = 1;
 var GETTER_MESSAGE_INDEX = 2;
 var ARG_LOCAL_LIMIT_MESSAGE_INDEX = 3;
-var STRING_LIMIT_MESSAGE_INDEX = 4;
 
 /**
  * Captures the stack and current execution state.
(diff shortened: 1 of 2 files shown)


src/agent/state.js
```diff
@@ -473,7 +473,8 @@ StateResolver.prototype.resolveMirror_ = function(mirror) {
   if (truncate) {
     members.push({name: 'Only first `config.capture.maxProperties=' +
                         this.config_.capture.maxProperties +
-                        '` properties were captured'});
+                        '` properties were captured. Use in an expression' +
+                        ' to see all properties.'});
```
ofrobots: Can you also fix the message for `STRING_LIMIT_MESSAGE_INDEX`?

test/test-v8debugapi.js
```diff
@@ -733,6 +733,7 @@ describe('v8debugapi', function() {
           assert(item.status.description.format.indexOf('Only first') !== -1);
           assert(item.status.description.format.indexOf(
             'config.capture.maxStringLength=3') !== -1);
+          assert(item.status.description.format.indexOf('of length 11.') !== -1);
```
ofrobots: nit: this would be more readable as a regex?

src/agent/state.js
matthewloring: Done in the latest set of revisions.

test/test-v8debugapi.js
matthewloring: Definitely, I'll put together a follow on to fix this throughout the file.

PR #271, opened 2017-06-06
include variables from outer scopes

ofrobots:
We were previously omitting variables from closure scopes. This can be
problematic specifically in transpiled async functions as most actual
variables end up in a closure.

### DIFF
--- a/src/agent/state.js
+++ b/src/agent/state.js
@@ -155,8 +155,8 @@ StateResolver.prototype.capture_ = function() {
     });
   }
 
-  // The frames are resolved after the evaluated expressions so that 
-  // evaluated expressions can be evaluated as much as possible within 
+  // The frames are resolved after the evaluated expressions so that
+  // evaluated expressions can be evaluated as much as possible within
   // the max data size limits
   var frames = that.resolveFrames_();
 
@@ -221,7 +221,7 @@ StateResolver.prototype.resolveFrames_ = function() {
   var frames = [];
   var frameCount = Math.min(this.state_.frameCount(),
     this.config_.capture.maxFrames);
-    
+
   for (var i = 0; i < frameCount; i++) {
     var frame = this.state_.frame(i);
     if (this.shouldFrameBeResolved_(frame)) {
@@ -344,7 +344,7 @@ StateResolver.prototype.extractArgumentsList_ = function (frame) {
 
 /**
  * Iterates and returns variable information for all scopes (excluding global)
- * in a given frame. FrameMirrors should return their scope object list with 
+ * in a given frame. FrameMirrors should return their scope object list with
  * most deeply nested scope first so variables initially encountered will take
  * precedence over subsequent instance with the same name - this is tracked in
  * the usedNames map. The argument list given to this function may be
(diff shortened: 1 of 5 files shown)


src/agent/state.js
```diff
@@ -361,17 +361,29 @@ StateResolver.prototype.resolveLocalsList_ = function (frame, args) {
...
+
+  // There will always be at least 3 scopes.
+  // For top-level breakpoints: [local, script, global]
+  // Other: [..., closure (module IIFE), script, global]
+  assert(count >= 3);
+  assert(allScopes[count - 1].scopeType() == ScopeType.Global);
+  assert(allScopes[count - 2].scopeType() == ScopeType.Script);
+
+  // We find the top-level (module global) variable pollute the local variables
+  // we we omit them by default, unless the breakpoint itself is top-level. The
```
matthewloring: Duplicate "we"

test/test-v8debugapi.js
```diff
@@ -1171,6 +1171,33 @@ describe('v8debugapi', function() {
...
+        }
+      };
+
+      const run = require('./fixtures/ts/async.js');
+      api.set(bp, (err) => {
+        assert.ifError(err);
+        process.nextTick(run);
+        api.wait(bp, (err) => {
+          assert.ifError(err);
+          // console.log(require('util').inspect(bp.stackFrames, {depth: null}));
```
matthewloring: Comment.

src/agent/state.js
```diff
@@ -361,17 +361,29 @@ StateResolver.prototype.resolveLocalsList_ = function (frame, args) {
...
   var usedNames = {};
   var makeMirror = this.ctx_.MakeMirror;
-  return flatten(frame.allScopes().map(
+  const allScopes = frame.allScopes();
+  const count = allScopes.length;
+
+  // There will always be at least 3 scopes.
+  // For top-level breakpoints: [local, script, global]
+  // Other: [..., closure (module IIFE), script, global]
+  assert(count >= 3);
```
matthewloring: It would be great if these could be `DCHECK`s. There are probably other sanity `DECHECK`s we could benefit from in this file.

src/agent/state.js
```diff
@@ -361,17 +361,29 @@ StateResolver.prototype.resolveLocalsList_ = function (frame, args) {
...
+  assert(allScopes[count - 1].scopeType() == ScopeType.Global);
+  assert(allScopes[count - 2].scopeType() == ScopeType.Script);
```
matthewloring: `===`

src/agent/state.js
ofrobots: I think this is a good idea but I think it would better suited for a follow-on.

test/test-v8debugapi.js
ofrobots: Done.

src/agent/state.js
ofrobots: Done.

src/agent/state.js
ofrobots: Done.

test/test-v8debugapi.js
```diff
@@ -1171,6 +1171,32 @@ describe('v8debugapi', function() {
...
+        location: {
+          path: path.join('.', 'test', 'fixtures', 'ts', 'async.js'),
+          line: 71
+        }
+      };
+
+      const run = require('./fixtures/ts/async.js');
+      api.set(bp, (err) => {
+        assert.ifError(err);
+        process.nextTick(run);
```
matthewloring: Placing this after the `api.wait` will be more consistent with the other tests and more clear about the desired order of execution.

ofrobots: The point about consistency is a good one. I do think this way is more desirable as the other way has an async race. I will make it consistent with the other tests for now though.

PR #272, opened 2017-06-06
disableable assertions

ofrobots:
state.js needs some assertions because it interfaces with parts of V8
debug API that are volatile across versions. However, we don't want to
be running these assertions in real user code. debug-assert module
allows us to run with assertions enabled in the CI.


bin/run-test.sh
```diff
@@ -2,6 +2,9 @@
 
 # Usage: -c to report coverage
 
+# Enable assertions
+export DEBUG_AGENT_ASSERTIONS=1
```
matthewloring: We've used the structure `CLOUD_DEBUG_xxxx` for our other env vars. Might be good to keep things consistent.

test/test-debug-assert.js
```diff
@@ -0,0 +1,34 @@
...
+
+describe('debug-assert', () => {
+  const debugAssert = require('../src/agent/debug-assert.js');
+
+  it('should not fire assertions when disabled', () => {
+    const assert = debugAssert(false);
+    assert.equal(1, 2);
+  });
+
+  it('should fire assertions when enabled', (done) => {
```
matthewloring: Could we add a test to ensure full assert api coverage?

bin/run-test.sh
ofrobots: Done.

test/test-debug-assert.js
ofrobots: Done.

Added another type of test to ensure that CI is running with assertions enabled.

matthewloring:
I think we also need the environment variable set on mocha.

ofrobots:
> I think we also need the environment variable set on mocha.

Yes. Users running mocha manually (rather than run-test.sh) will need to make sure that the env. var. is defined. The purpose of the test is the make sure that our CI always runs with the env. var. is defined.

I am not sure if you are requesting a change or making an observation. Can you approve or elaborate on what change you would like to see?

matthewloring:
Complete mistype. Meant to say appveyor. Should the environment variable be set on appveyor?

ofrobots:
Ah, so appveyor doesn't use run-test.sh? Ofcourse it doesn't. Dx

ofrobots:
PTAL.

PR #274, opened 2017-06-07
Use FUNCTION_NAME as service id on GCF

matthewloring:


matthewloring:
Before: 
screen shot 2017-06-06 at 5 56 55 pm

--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
After:
screen shot 2017-06-06 at 6 59 14 pm

ofrobots:
@swalkowski: Is there a concept of 'versions' for functions?

src/agent/debuglet.js
```diff
@@ -146,8 +146,10 @@ Debuglet.normalizeConfig_ = function(config) {
   var envConfig = {
     logLevel: process.env.GCLOUD_DEBUG_LOGLEVEL,
     serviceContext: {
-      service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
-      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+      service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME ||
+        process.env.FUNCTION_NAME,
+      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION ||
+        'unversioned',
```
ofrobots: When the app is running locally, we will be providing a serviceContext with `service` set to undefined and `version` set to `unversioned`. How would this display in the UI?

matthewloring: It should display the command used to start the process as is done today.

swalkowski:
@ofrobots: No, functions are unversioned at the moment.

src/agent/debuglet.js
```diff
@@ -146,8 +146,10 @@ Debuglet.normalizeConfig_ = function(config) {
   var envConfig = {
     logLevel: process.env.GCLOUD_DEBUG_LOGLEVEL,
     serviceContext: {
-      service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
-      version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+      service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME ||
```
DominicKramer: Is there a known situation where `GAE_SERVICE` or `GAE_MODULE_NAME` are intentionally set to the empty string.  I don't think this is the case, but I just want to make sure.

src/agent/debuglet.js
matthewloring: Whoops, looks like that was not the case. Fixed.

src/agent/debuglet.js
matthewloring: Not to my knowledge.

PR #277, opened 2017-06-12
Document the support of transpiled code

DominicKramer:


README.md
```diff
@@ -138,6 +138,33 @@ As soon as that line of code is reached in any of the running instances of your
 
 ![Breakpoint Hit](doc/images/breakpoint-hit.png?raw=true)
 
+Note that the directory layout of the code that is being debugged does not have to exactly match the source code specified in the Debug UI.  In particular, if a snapshot is specified in a file in the Debug UI, a breakpoint is set in the corresponding file in the code of the running instance that matches the most path components of the file specified in the Debug UI starting from the rightmost path component and working left.
```
matthewloring: maybe "...a breakpoint is set in the deployed file with the longest matching path suffix" would be more clear?

README.md
```diff
@@ -138,6 +138,33 @@ As soon as that line of code is reached in any of the running instances of your
...
+
+An example will help illustrate this.  That is, suppose that the code on the running instance is structured as follows:
+```
+/running/instance/main.js
+/running/instance/a/index.js
+/running/instance/b/a/index.js
+```
+Now suppose on your device you have a copy of the code as follows, and you have selected the local `/your/device` directory in the Debug UI as the directory containing the code to debug.
```
matthewloring: maybe "local machine" is a better fit than "device"?

README.md
```diff
@@ -138,6 +138,33 @@ As soon as that line of code is reached in any of the running instances of your
...
+```
+/your/device/main.js
+/your/device/a/index.js
+/your/device/b/a/index.js
+```
+
+In this case, if you specify a snapshot in file `/your/device/b/a/index.js` in the Debug UI, the debugger will identify that that file corresponds to the file `/running/instance/b/a/index.js` of the code on the running instance, and the breakpoint will hit when the specified line of `/running/instance/b/a/index.js` is reached.
+
+Note, however, if a snapshot is specified for the file `/your/device/a/index.js` in the Debug UI, then the debugger would not know whether this file corresponds to the file `/running/instance/a/index.js` or `/running/instance/b/a/index.js`.  If such an ambiguity occurs, a message will be displayed in the Debug UI.
```
matthewloring: It would be good to add a note about how `config#appPathRelativeToRepository` can help resolve these ambiguities.

README.md
```diff
@@ -138,6 +138,33 @@ As soon as that line of code is reached in any of the running instances of your
...
+
+## Support for Transpiled Code
+
+The debugger supports the use of transpiled code, whether it be Javascript to Javascript transpilation or transpilation from another language (such as Typescript) to Javascript.  In order to use transpiled code with the debugger, sourcemaps need to be generated with the transpiled code.  The sourcemaps can either be included inline in the transpiled code or provided in separate `.js.map` files.
+
+To use the debugger, the only files that are needed at runtime are the transpiled files as well as the generated sourcemap files (if the sourcemaps are not inlined).  In particular, the original source does not need to be available to the debugger at runtime.
```
matthewloring: maybe "...only files that are needed in the deployment environment are the transpiled..."

README.md
```diff
@@ -138,6 +138,33 @@ As soon as that line of code is reached in any of the running instances of your
...
+/your/device/b/a/index.js
+```
+
+In this case, if you specify a snapshot in file `/your/device/b/a/index.js` in the Debug UI, the debugger will identify that that file corresponds to the file `/running/instance/b/a/index.js` of the code on the running instance, and the breakpoint will hit when the specified line of `/running/instance/b/a/index.js` is reached.
+
+Note, however, if a snapshot is specified for the file `/your/device/a/index.js` in the Debug UI, then the debugger would not know whether this file corresponds to the file `/running/instance/a/index.js` or `/running/instance/b/a/index.js`.  If such an ambiguity occurs, a message will be displayed in the Debug UI.
+
+## Support for Transpiled Code
+
+The debugger supports the use of transpiled code, whether it be Javascript to Javascript transpilation or transpilation from another language (such as Typescript) to Javascript.  In order to use transpiled code with the debugger, sourcemaps need to be generated with the transpiled code.  The sourcemaps can either be included inline in the transpiled code or provided in separate `.js.map` files.
```
matthewloring: Could you double check that inline source maps are supported?

DominicKramer: It doesn't look like inline sourcemaps are supported and so I removed this for now.

README.md
DominicKramer: That sounds better.  I've updated the docs.

README.md
DominicKramer: Done.

README.md
DominicKramer: I like "local machine" better and updated the docs.

README.md
DominicKramer: That does sound better.  I've updated the docs to use it.

README.md
```diff
@@ -138,6 +138,37 @@ As soon as that line of code is reached in any of the running instances of your
...
+An example will help illustrate this.  That is, suppose that the code on the running instance is structured as follows:
+```
+/running/instance/package.json
+/running/instance/main.js
+/running/instance/a/index.js
+/running/instance/b/a/index.js
+```
+Now suppose on your local machine you have a copy of the code as follows, and you have selected the `/local/machine` directory in the Debug UI as the directory containing the code to debug.
+```
+/local/machine/package.json
```
ofrobots: It would be better to talk about repo paths rather than absolute paths so that this applies more generally to source in github etc. Concretely, I would recommend omitting the `/local/machine` from the prefix here and talking about 'Now suppose that your source code repository is structured as follow...'

README.md
```diff
@@ -138,6 +138,37 @@ As soon as that line of code is reached in any of the running instances of your
...
+/local/machine/main.js
+/local/machine/a/index.js
+/local/machine/b/a/index.js
+```
+
+In this case, if you specify a snapshot in file `/local/machine/b/a/index.js` in the Debug UI, the debugger will identify that that file corresponds to the file `/running/instance/b/a/index.js` of the code on the running instance, and the breakpoint will hit when the specified line of `/running/instance/b/a/index.js` is reached.
```
ofrobots: My understanding is that snapshot paths are relative to source root; so it would be `b/a/index.js` and it would not contain the prefix.

README.md
```diff
@@ -138,6 +138,37 @@ As soon as that line of code is reached in any of the running instances of your
...
+
+Note, however, if a snapshot is specified for the file `/local/machine/a/index.js` in the Debug UI, then the debugger would not know whether this file corresponds to the file `/running/instance/a/index.js` or `/running/instance/b/a/index.js`.  If such an ambiguity occurs, a message will be displayed in the Debug UI.
+
+In this case, the `appPathRelativeToRepository` configuration option can be used to specify the directory containing the application's `package.json` file to help resolve the ambiguity.  In the example above, if the `appPathRelativeToRepository` configuration option is set to `/local/machine`, the debugger would be able to identify `/running/instance/a/index.js` as the file where the breakpoint should be set.  See the [Debugger Agent Settings](#debugger-agent-settings) section for more information on configuring the debugger.
```
ofrobots: Is that how `appPathRelativeToRepostiory` works? What happens when the source is on github? The name sounds wrong if `appPathRelativeToRepository` is an absolute path.

README.md
DominicKramer: Updated.

README.md
DominicKramer: I've verified that the snapshots are relative, and I've updated the text.

README.md
DominicKramer: I think the `appPathRelativeToRepository` config option would be best documented more thoroughly in a follow-up PR.

README.md
```diff
@@ -138,6 +138,35 @@ As soon as that line of code is reached in any of the running instances of your
 
 ![Breakpoint Hit](doc/images/breakpoint-hit.png?raw=true)
 
+Note that the directory layout of the code that is being debugged does not have to exactly match the source code specified in the Debug UI.  In particular, if a snapshot is specified in a file in the Debug UI, a breakpoint is set in the deployed file with the longest matching path suffix.
```
ofrobots: copy editing nits: 

* highlight *not* in first sentence.
* second sentence would read better like this: "The agent resolves snapshot filenames by searching for a file with the longest matching path suffix. If a use unique match is found, we will use that to set the snapshot.

README.md
```diff
@@ -138,6 +138,35 @@ As soon as that line of code is reached in any of the running instances of your
...
+
+Note, however, if a snapshot is specified for the file `a/index.js` in the Debug UI, then the debugger would not know whether this file corresponds to the file `/running/instance/a/index.js` or `/running/instance/b/a/index.js`.  If such an ambiguity occurs, a message will be displayed in the Debug UI.
+
+## Support for Transpiled Code
+
+The debugger supports the use of transpiled code, whether it be Javascript to Javascript transpilation or transpilation from another language (such as Typescript) to Javascript.  In order to use transpiled code with the debugger, sourcemaps need to be generated with the transpiled code.  The sourcemaps need to be provided in `.js.map` files.
+
+To use the debugger, the only files that are needed in the deployment environment are the transpiled files as well as the generated sourcemap files.  In particular, the original source does not need to be available to the debugger in the deployment environment.
```
ofrobots: copy edit:

"You do not need to deploy the original source files to the deployment environment as long as the sourcemap files are available at runtime"

README.md
```diff
@@ -138,6 +138,35 @@ As soon as that line of code is reached in any of the running instances of your
...
+
+Instead, the original source code is provided to the Debug UI to specify snapshots.  When specifying a snapshot in an original source file in the Debug UI, the corresponding file and line in the transpiled code is automatically determined based on the sourcemap files provided with the transpiled code at runtime.  See the [Using the Debugger](#using-the-debugger) section for more information about using the Debug UI.  In addition, the exact directory layout of the original source is somewhat flexible, just as it is with the use of non-transpiled code as described in the [Using the Debugger](#using-the-debugger) section.
```
ofrobots: copy edit: "In the Debug UI, you only need to provide the original source code – you don't need the transpiled output files or the sourcemaps. When you set a snapshot in an original source file in the...."

README.md
```diff
@@ -138,6 +138,35 @@ As soon as that line of code is reached in any of the running instances of your
...
+```
+package.json
+main.js
+a/index.js
+b/a/index.js
+```
+
+In this case, if you specify a snapshot in file `b/a/index.js` in the Debug UI, the debugger will identify that that file corresponds to the file `/running/instance/b/a/index.js` of the code on the running instance, and the breakpoint will hit when the specified line of `/running/instance/b/a/index.js` is reached.
+
+Note, however, if a snapshot is specified for the file `a/index.js` in the Debug UI, then the debugger would not know whether this file corresponds to the file `/running/instance/a/index.js` or `/running/instance/b/a/index.js`.  If such an ambiguity occurs, a message will be displayed in the Debug UI.
```
kjin: In the case of ambiguity, does the UI default to one of these options?

README.md
DominicKramer: Updated.

README.md
DominicKramer: Updated.

README.md
DominicKramer: Updated.

README.md
DominicKramer: In this case, the snapshot is not set.  I've updated the text to make that more clear.

PR #275, opened 2017-06-08
Use service name as debuggee id on gke

matthewloring:

PR #278, opened 2017-06-19
Revert "Use service name as debuggee id on gke (#275)"

ofrobots:
This reverts commit d3994f89599258d93c4d887ad5a605fb005ba168 as it breaks on non-GKE GCP enviornments.

See: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/275#issuecomment-309308717


ofrobots:
TBR= @matthewloring

PR #281, opened 2017-06-23
Initial gulpfile that only maps src.ts to src

DominicKramer:


gulpfile.js
```diff
@@ -0,0 +1,66 @@
...
+});
+
+gulp.task('test.unit', ['compile'], cb => {
+  spawn('bash', ['./bin/run-test.sh'], {
+    stdio : 'inherit'
+  }).on('close', cb);
+});
+
+gulp.task('test', ['test.unit']);
+gulp.task('default', ['compile']);
```
kjin: End file with newline here and in the `ts*.json` files

package.json
```diff
@@ -45,11 +54,11 @@
     "split": "^1.0.0"
   },
   "scripts": {
-    "test": "./bin/run-test.sh",
-    "system-test": "./bin/run-system-test.sh",
-    "changelog": "./bin/run-changelog.sh",
-    "coverage": "./bin/run-test.sh -c",
-    "bump": "./bin/run-bump.sh",
-    "closure": "./node_modules/.bin/closure-npc"
+    "test": "gulp test",
```
ofrobots: It seems that local testing by developers will be different from what travis runs. Intentional? if so, why?

.travis.yml
```diff
@@ -5,7 +5,7 @@ node_js:
 - '6'
 - '8'
 script:
-- ./bin/run-test.sh -c
+- gulp && ./bin/run-test.sh -c
```
JustinBeckwith: what version of gulp would that run?  Shouldn't this be npm run build or npx gulp?

appveyor.yml
```diff
@@ -18,7 +18,7 @@ install:
 # Post-install test scripts.
 test_script:
   # run tests
-  - ps: node_modules/.bin/mocha test --timeout 4000 --R
+  - ps: node_modules/.bin/gulp ; node_modules/.bin/mocha test --timeout 4000 --R
```
JustinBeckwith: wouldn't we want npm scripts here?

package.json
```diff
@@ -45,11 +54,11 @@
     "split": "^1.0.0"
   },
   "scripts": {
```
JustinBeckwith: no new @ types?

appveyor.yml
ofrobots: I think that would be a good improvement, but it would be good to limit this PR to a refactor. There are 7 more coming after this.

package.json
DominicKramer: The master branch has the same differences.  That is, local testing doesn't do coverage, but coverage is done on travis.  Since this PR is consistent with master in terms of behavior, I think addressing this difference can be done after the transition to typescript is complete.

appveyor.yml
DominicKramer: I agree we would want npm scripts here.  The code in the master branch doesn't use npm scripts either, and I am trying not to change behavior in the translation to typescript PRs.  I'm concerned that perhaps there is a subtle reason why npm scripts aren't being used, and thus it could be better to add npm scripts in a follow-up PR that is independent of the translation to typescript.

.travis.yml
DominicKramer: Good catch.  I've added a `build` script and now `npm run build` is run instead of directly running `gulp`.

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": false,
+    "allowUnreachableCode": false,
+    "allowUnusedLabels": false,
+    "alwaysStrict": true,
+//    "declaration": true,
```
matthewloring: Comments are not part of the json spec. Not sure if it happens to work in this case.

ofrobots: It is supported by the typescript compiler.

DominicKramer: Also one of the PRs I will open later will remove all of these comments to enable all options.

appveyor.yml
ofrobots: As long as there is a list you're keeping of the follow-ons, sgtm.

PR #285, opened 2017-06-27
Use Typescript class syntax

DominicKramer:


src.ts/agent/debuglet.js
```diff
@@ -30,7 +30,7 @@ var common = require('@google-cloud/common');
 var v8debugapi = require('./v8debugapi.js');
 var Debuggee = require('../debuggee.js');
 var DebugletApi = require('../controller.js');
-var defaultConfig = require('./config.js');
+var defaultConfig = require('./config.js').default;
```
ofrobots: nit: `require('./config').default`

DominicKramer: This will be addressed in a follow-up PR when `import` syntax is introduced.

PR #287, opened 2017-06-28
typo: wait_expired -> waitExpired

ofrobots:
This was causing breakpoints to be locally expired incorrectly.

### DIFF
--- a/system-test/test-e2e.js
+++ b/system-test/test-e2e.js
@@ -129,7 +129,7 @@ describe('@google-cloud/debug end-to-end behavior', function () {
   });
 
   it('should set breakpoints correctly', function() {
-    this.timeout(25 * 1000);
+    this.timeout(90 * 1000);
     // Kick off promise chain by getting a list of debuggees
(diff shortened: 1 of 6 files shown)


ofrobots:
Adjusted the e2e test to better exercise waitExpired. Also fix `success_on_timeout` -> `successOnTimeout`.

system-test/test-e2e.js
```diff
@@ -197,13 +197,13 @@ describe('@google-cloud/debug end-to-end behavior', function () {
           'transcript in child ' + index + ' should contain value of o: ' +
           child.transcript);
       });
-      return api.deleteBreakpoint(debuggeeId, breakpoint.id);
+      return Promise.resolve();
```
DominicKramer: Is the `value` argument to `Promise.resolve()` optional?

ofrobots: Yep, it can be omitted. The promise will be settled with an `undefined` value.

PR #289, opened 2017-06-28
Convert `require`s to `import`s

DominicKramer:


src.ts/agent/scanner.ts
```diff
@@ -213,7 +209,8 @@ function stats(filename, shouldHash, cb: (err, stats?: FileStats) => void) {
   if (shouldHash) {
     shasum = crypto.createHash('sha1');
   }
-  var s = fs.ReadStream(filename);
+  // TODO: Determine why property 'ReadStream' does not exist on type 'fs'
+  var s = (fs as any).ReadStream(filename);
```
ofrobots: I think the intention was to use `fs.createReadStream`? I think it is fine to keep this as a TODO and fix after the refactors are done.

DominicKramer: Sounds good.

PR #290, opened 2017-06-28
Convert `var` to `const` or `let`

DominicKramer:


src.ts/agent/v8debugapi.ts
```diff
@@ -22,7 +22,7 @@ import * as semver from 'semver';
 import * as state from './state';
 import { StatusMessage } from '../status-message';
 
-/** @const */ var messages = {
+/** @const */ const messages = {
```
ofrobots: nit: get rid of the `@ const` annotation.

DominicKramer: Done.

PR #291, opened 2017-06-28
Add types to all non-test files

DominicKramer:


src.ts/agent/debuglet.ts
```diff
@@ -256,16 +262,17 @@ export class Debuglet extends EventEmitter {
   /**
    * @private
    */
-  static createDebuggee(projectId, uid, serviceContext, sourceContext, description,
-                        errorMessage, onGCP) {
+  // TODO: Determine the type of sourceContext
+  static createDebuggee(projectId: string, uid: string, serviceContext: { service?: string, version?: string, minorVersion_?: string }, sourceContext: { [key: string]: string }, description: string,
```
DominicKramer: I think I added that in later PRs.

src.ts/agent/debuglet.ts
```diff
@@ -179,7 +185,7 @@ export class Debuglet extends EventEmitter {
    * as it is on the critical path of application startup.
    * @private
    */
-  start() {
+  start(): void {
```
DominicKramer: I added return types on everything to make sure what I thought the return should be agreed with what was inferred.

PR #293, opened 2017-06-29
Enable all `tsconfig.json` options

DominicKramer:


src.ts/agent/scanner.ts
```diff
@@ -15,20 +15,36 @@
  */
 
 import * as crypto from 'crypto';
-import * as findit from 'findit2';
+
+import * as events from 'events';
+const _findit = require('findit2');
```
ofrobots: nit: would the following work?

```js
const findit: (dir: string) => events.EventEmitter = require('findit2');
```
If so, can you use this pattern elsewhere as well?

DominicKramer: That looks much better.  It works, and I've updated the code.

PR #295, opened 2017-06-29
Update `debug.ts` to use class syntax

DominicKramer:


src.ts/debug.ts
```diff
@@ -59,7 +52,12 @@ export class Debug {
     };
 
     // TODO: Update Service to provide types
-    common.Service.call(this, config, options);
+    super(config, options || {});
```
ofrobots: It seems awkward to call `super` before we have determined that this function was called as a constructor. Since `this` is not available before the call to `super`, the `instanceof` check cannot be before the call to `super`. Suggest using `new.target` instead of `instanceof`.

DominicKramer: I've updated the code to use `new.target`.

DominicKramer: I've updated the code to move the `normalizeArguments` call after the call to `super`.  PTAL

PR #296, opened 2017-06-30
Build and packaging cleanups

ofrobots:
* Add a files section to the package.json to restrict the files that get shipped as part of the package.
* No more .js files in the src.ts directory
* Produce output files in `build/src` and `build/types`.
* Use correct sourceRoot for sourcemaps
* Run mocha with source map support.

It would be easiest to review commits chronologically.


gulpfile.js
```diff
@@ -31,8 +31,8 @@ const format = require('gulp-clang-format');
 
 const tsconfigPath = path.join(__dirname, 'tsconfig.json');
 const tslintPath = path.join(__dirname, 'tslint.json');
-const outDir = '.';
-const sources = ['src.ts/**/*.ts', 'src.ts/**/*.js'];
+const outDir = 'build';
+const sources = ['src.ts/**/*.ts'];
```
DominicKramer: If the output is now placed in `build/src`, we can rename `src.ts` back to `src`, but it could be better to do that in a follow-up.

ofrobots: yep - that plan is to do this in a follow-up.

PR #300, opened 2017-07-06
tune up configuration types

ofrobots:
* Get rid of jsdoc types from comments – not needed now that we TS types. Not entirely sure yet how jsdoc documentation generator handles ts, but we shall find out.
* Proper inheritance for Auth options.
* Fix some TODOs

### DIFF
--- a/src/debug.ts
+++ b/src/debug.ts
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-import {AuthOptions, Common} from './types/common-types';
+import {AuthenticationConfig, Common} from './types/common-types';
 export const common: Common = require('@google-cloud/common');
 
 export class Debug extends common.Service {
-  options: AuthOptions;
+  options: AuthenticationConfig;
 
   /**
    * <p class="notice">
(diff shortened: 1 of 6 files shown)


src/debug.ts
```diff
@@ -43,7 +43,7 @@ export class Debug extends common.Service {
    *
    * @param {object} options - [Configuration object](#/docs)
```
DominicKramer: I think we should update the type here or take it out.

src/index.ts
```diff
@@ -27,7 +26,6 @@ let debuglet: Debuglet;
  * with Stackdriver Debug.
  *
  * @param {object=} options - Options
```
DominicKramer: I think we should update the type here or not specify a type.

src/debug.ts
ofrobots: done.

src/index.ts
ofrobots: done.

src/index.ts
```diff
@@ -26,31 +25,27 @@ let debuglet: Debuglet;
...
-  const agentConfig = (options as any).debug || options;
+  const agentConfig: DebugAgentConfig =
+      (options as StackdriverConfig).debug || (options as DebugAgentConfig);
 
   // forceNewAgent_ is for testing purposes only.
   if (debuglet && !agentConfig.forceNewAgent_) {
     throw new Error('Debug Agent has already been started');
   }
 
   // TODO: Determine how to remove this cast to `AuthOptions`.
```
DominicKramer: It looks like this TODO can be removed.

ofrobots: Good catch.

PR #301, opened 2017-07-06
Lint and yarn

ofrobots:


src/agent/state.ts
```diff
@@ -154,7 +154,7 @@ class StateResolver {
     // Evaluate the watch expressions
     const evalIndexSet = new Set();
     if (that.expressions_) {
-      that.expressions_.forEach(function(expression, index) {
+      that.expressions_.forEach(function(expression, index2) {
```
DominicKramer: It could be helpful to have the first occurrence of `index` changed to `index1` too.

src/agent/scanner.ts
```diff
@@ -207,7 +207,7 @@ function findFiles(
     return;
   });
 
-  find.on('directory', function(dir: string, _: fs.Stats, stop: () => void) {
+  find.on('directory', function(dir: string, _ignore: fs.Stats, stop: () => void) {
```
DominicKramer: We could use `__` here too if we wanted.

src/agent/debuglet.ts
```diff
@@ -392,7 +392,7 @@ export class Debuglet extends EventEmitter {
     //       named `response`) is not used.
     metadata.project(
         'project-id',
-        function(err: Error, _: http.ServerResponse, metadataProject: string) {
+        function(err: Error, _res: http.ServerResponse, metadataProject: string) {
```
DominicKramer: The current linting setup allows leading underscores, but we should determine if we want leading and/or trailing underscores.  In the `StateResolver` class we also use trailing underscores.  We should determine a standard approach, but we can do that in a follow-up PR.

ofrobots: There are two distinct uses of underscores.. in parameter names, leading underscores indicate to the typescript compiler that the property may be unused. For property names, they primarily indicate to JS users of the library that the property is intended to be private.

PR #304, opened 2017-07-07
lint, tooling, and other misc. changes

ofrobots:
* Remove jshint files and devDependency
* Update tslint config. Use typechecked lint. Remove options that are subsumed
  by the compiler now.
* Start shipping type definitions.
* Address newly found lint errors.
* Formatting.


ofrobots:
Thanks, removed the commented out jshint invocation from run-test.sh

PR #307, opened 2017-07-10
Start using async/await to flatten the pyramid in Debuglet.start

ofrobots:
Start using async/await to simplify Debuglet.start. This exercise can be continued for the remainder of the nested callbacks in follow-on PRs. This PR peels the first two layers only.

For an easier time reviewing, review each commit individually.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -220,85 +220,80 @@ export class Debuglet extends EventEmitter {
     if (process.env.GAE_MINOR_VERSION) {
       id = 'GAE-' + process.env.GAE_MINOR_VERSION;
     }
-    // TODO: Address the case when `that.config_.workingDirectory` is
-    //       `null`.
-    scanner.scan(
-        !id, that.config_.workingDirectory as string, /.js$|.map$/,
-        function(
-            err2: Error|null, fileStats?: scanner.ScanResults, hash?: string) {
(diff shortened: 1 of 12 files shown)


ofrobots:
Also.. adding `?w=1` to the URL makes the code a lot easier to review. Why isn't this the default on github?

src/agent/debuglet.ts
```diff
@@ -198,107 +200,100 @@ export class Debuglet extends EventEmitter {
...
-                      that.emit('started');
-                    });
-                  });
-                });
-              });
+    const jsStats = fileStats.selectStats(/.js$/);
+    const mapFiles = fileStats.selectFiles(/.map$/, process.cwd());
+    SourceMapper.create(mapFiles, function(err3, mapper) {
+      if (err3) {
+        that.logger_.error('Error processing the sourcemaps.', err3);
```
DominicKramer: nit: It looks like `err3` can be renamed to `err1` now.

ofrobots: That would have inflated the patch without having any real meat to the change. The real fix would be to flatten that part of the code as well.

I think we should keep the variable as err3 to avoid unnecessary churn.

DominicKramer: Sounds good.  The name isn't really an issue.

PR #308, opened 2017-07-11
Simplify metadata

ofrobots:
The first commit adds a simpler, more testable version of code to lookup metadata. The second commit uses this. Review each commits separately.

Note that we no longer query the metadata service if projectId is locally available.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -20,6 +20,7 @@ const common: Common = require('@google-cloud/common');
 import * as crypto from 'crypto';
 import {EventEmitter} from 'events';
 import * as extend from 'extend';
+import * as dns from 'dns';
 import * as fs from 'fs';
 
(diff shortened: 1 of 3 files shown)


src/agent/debuglet.ts
```diff
@@ -405,6 +406,38 @@ export class Debuglet extends EventEmitter {
...
+    }
+    return project;
+  }
+
+  static async runningOnGCP(): Promise<boolean> {
+    const lookup = promisify(dns.lookup);
+    try {
+      await lookup('metadata.google.internal.');
+      return true;
+    } catch (err) {
```
DominicKramer: We should probably have a comment here describing why it is ok to ignore `err` so that, looking back, we know it was intentional.

ofrobots: Done.

Issue #280, opened 2017-06-19
Re-implement service naming for GKE


ofrobots:
This was reverted in 66d67f8b1bdc26138f29ed29988d688e6b941853.

PR #309, opened 2017-07-12
Use clusterName as service name on GKE

ofrobots:
Fixes: #280

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -402,6 +402,16 @@ export class Debuglet extends EventEmitter {
     });
   }
 
+  static getClusterNameFromMetadata(): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/agent/debuglet.ts
```diff
@@ -262,6 +262,24 @@ export class Debuglet extends EventEmitter {
         return;
       }
 
+      if (onGCP &&
+          (!that.config_.serviceContext ||
+           !that.config_.serviceContext.service)) {
```
DominicKramer: If the user specifies the empty string as the `service` we will still treat it as unspecified.  However, I don't think that is necessarily an issue.

ofrobots: Ack. I think the behaviour of fetching the cluster name on an empty string sounds reasonable still, so I agree this is not an issue.

PR #313, opened 2017-07-19
Move `fixtures` directory to the project root

DominicKramer:


gulpfile.js
```diff
@@ -90,7 +92,7 @@ gulp.task('test.compile', ['compile'], () => {
 });
 
 gulp.task('test.unit', ['compile'], cb => {
-  spawn('bash', ['./bin/run-test.sh'], {
+  spawn('bash', ['./bin/run-test.sh'].concat(taskArgs), {
```
matthewloring: It may be good to have a more general mechanism for arg parsing if we end up with many gulp tasks that require arguments. Since there is currently only one task which has one optional argument, maybe it would be cleaner to have separate gulp tasks for `test.unit` and `coverage`?

DominicKramer:
This has been replaced with https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/314.

PR #315, opened 2017-07-28
Update test files to use the `.ts` extension

DominicKramer:


test/nocks.ts
```diff
@@ -13,16 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-'use strict';
 
-var nock = require('nock');
+ var nock = require('nock');
```
ofrobots: nit: unnecessary whitespace. It is okay if it is fixed in a follow-on (if fixing this requires that you have to rebase all your pending PRs).

test/test-duplicate-expressions-code.js
```diff
@@ -0,0 +1,8 @@
+/*1* KEEP THIS CODE AT THE TOP TO AVOID LINE NUMBER CHANGES */
+/*2*/'use strict';
+/*3*/function foo(a) {
+/*4*/  process.nextTick(function() {
+/*5*/    a = 0;
+/*6*/  });
+/*7*/}
+/*8*/module.exports = foo;
```
ofrobots: nit: missing EOL

test/test-duplicate-nested-expressions-code.js
```diff
@@ -0,0 +1,11 @@
...
+/*2*/'use strict';
+/*3*/function foo(a) {
+/*4*/ var a = 10;
+/*5*/ a += 1;
+/*6*/ return (function (b) {
+/*7*/   var a = true;
+/*8*/   return a;
+/*9*/ }());
+/*10*/}
+/*11*/module.exports = foo;
```
ofrobots: missing EOL

test/test-max-data-size-code.js
```diff
@@ -0,0 +1,6 @@
+/*1* KEEP THIS CODE AT THE TOP TO AVOID LINE NUMBER CHANGES */
+/*2*/'use strict';
+/*3*/function foo(n) {
+/*4*/  var A = new Array(3); return n+42+A[0];
+/*5*/}
+/*6*/module.exports = foo;
```
ofrobots: nit: missing EOL

test/test-this-context-code.js
```diff
@@ -0,0 +1,14 @@
...
+/*5*/ this.a += b;
+/*6*/ return this;
+/*7*/}
+/*8*/function bar(j) {
+/*9*/ return j;
+/*10*/}
+/*11*/module.exports = {
+/*12*/  foo: foo,
+/*13*/  bar: bar
+/*14*/};
```
ofrobots: nit: missing EOL

test/test-try-catch-code.js
```diff
@@ -0,0 +1,11 @@
...
+/*2*/'use strict';
+/*3*/function foo() {
+/*4*/ try {
+/*5*/   throw new Error('A test');
+/*6*/ } catch (e) {
+/*7*/   var e = 2;
+/*8*/   return e;
+/*9*/ }
+/*10*/}
+/*11*/module.exports = foo;
```
ofrobots: ditto

test/test-v8debugapi-code.js
```diff
@@ -0,0 +1,14 @@
...
+/*5*/  return n+42+A[0]+B.b;
+/*6*/}
+/*7*/function getterObject() {
+/*8*/  var hasGetter = { _a: 5, get a() { return this._a; }, b: 'hello world' };
+/*9*/  return hasGetter.a;
+/*10*/}
+/*11*/module.exports = {
+/*12*/  foo: foo,
+/*13*/  getterObject: getterObject
+/*14*/};
```
ofrobots: ditto

tsconfig.json
```diff
@@ -3,19 +3,19 @@
     "allowSyntheticDefaultImports": false,
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
-    "alwaysStrict": true,
+//    "alwaysStrict": true,
```
ofrobots: I am guessing that these settings will be re-enabled when you're done refactoring? We'll need to make sure that no other changes land while you're working through the refactoring PRs. Is that understanding correct?

test/misc/bench-code.js
```diff
@@ -0,0 +1,6 @@
+/*1* KEEP THIS CODE AT THE TOP TO AVOID LINE NUMBER CHANGES */
```
kjin: I would suggest placing `-code.js` files in a different directory, since otherwise they might be confused with real tests at a glance.

test/test-v8debugapi.ts
```diff
@@ -26,12 +16,13 @@
 
 var breakpointInFoo = {
   id: 'fake-id-123',
-  location: { path: 'test-v8debugapi.js', line: 4 }
+  // TODO: Determine if we should be restricting to only the build directory.
+  location: { path: 'build/test/test-v8debugapi-code.js', line: 4 }
```
kjin: Why does only this test have `build/test` in its path?

test/nocks.ts
DominicKramer: This is a consequence of the auto indenting of the editor I was using.  A follow-up PR will replace this with `import * as nock from nock` and so this won't be an issue.

test/test-duplicate-expressions-code.js
DominicKramer: To prevent merge conflicts with landing the upcoming PRs, I won't fix this in this PR.  I'll record this so we don't forget to add EOL for all the `-code.js` files after the PRs have all landed.

tsconfig.json
DominicKramer: Yes that is correct.  There are nine PRs in total with the PRs at the end re-enabling the compiler options.

test/misc/bench-code.js
DominicKramer: I was thinking about doing that.  I think that is something that I can do in a follow-up PR after landing the Typescript PRs to avoid merge conflicts.

test/test-v8debugapi.ts
DominicKramer: I think this is because this test has a slightly different `workingDirectory` option set compared to the other tests.

PR #316, opened 2017-07-31
Update the tests to use `import` syntax

DominicKramer:


test/debugger.ts
```diff
@@ -20,10 +20,16 @@
...
-var util = require('util');
+import * as commonTypes from '../src/types/common-types';
+
+var pjson = require('../../package.json');
+const common: commonTypes.Common = require('@google-cloud/common');
+// TODO: Verify these types are correct.
+const qs: { parse: (qs: any, sep?: string, eq?: string,
+  options?: { maxKeys?: number }) => any,
+  stringify: (obj: object|string|boolean|number, sep?: string,
+    eq?: string, name?: string) => string} = require('querystring');
```
matthewloring: Shouldn't this type be available through the node typings module?

DominicKramer: Yes, the type should be available.  There is also a `querystring` module on `npm` that I thought this was in reference to (that didn't have type information).

I have a list of follow-up actions that I will complete after all of the Typescript related PRs have landed.  I will add this to that list, and I think it would be easiest to update this at that time to avoid merge conflicts with the upcoming PRs.

DominicKramer:
@matthewloring I tried to have each PR only change things that were actually needed to be changed for transpilation and tests to pass.  I think this will make it easier to review these changes as well as, in the future, review git history of the changes.  As a result, a lot of TODOs were introduced, but I think follow-up PRs to address them should be ok.

PR #324, opened 2017-08-08
Address TODOs

DominicKramer:
As part of the change to Typescript, many TODOs were added to the code for items to address.  This PR addresses some of these TODOs.

### DIFF
--- a/src/agent/config.ts
+++ b/src/agent/config.ts
@@ -17,12 +17,12 @@
 import {AuthenticationConfig} from '../types/common-types';
 
 export interface DebugAgentConfig extends AuthenticationConfig {
-  workingDirectory: string|null;
+  workingDirectory?: string;
 
   /**
    * A user specified way of identifying the service
    */
-  description: string|null;
+  description?: string;
 
   /**
    * Whether or not it is permitted to evaluate expressions.
@@ -44,17 +44,17 @@ export interface DebugAgentConfig extends AuthenticationConfig {
     /**
      * The service name.
      */
-    service: string | null;
+    service?: string;
 
     /**
      * The service version.
      */
-    version: string | null;
+    version?: string;
 
     /**
      * A unique deployment identifier. This is used internally only.
      */
-    minorVersion_: string | null;
+    minorVersion_?: string;
   };
 
   /**
@@ -65,7 +65,7 @@ export interface DebugAgentConfig extends AuthenticationConfig {
    * cases where the debug agent is unable to resolve breakpoint locations
    * unambiguously.
    */
-  appPathRelativeToRepository: string|null;
+  appPathRelativeToRepository?: string;
 
   /**
    * agent log level 0-disabled, 1-error, 2-warn, 3-info, 4-debug
@@ -170,14 +170,15 @@ const defaultConfig: DebugAgentConfig = {
   // FIXME(ofrobots): presently this is dependent what cwd() is at the time this
   // file is first required. We should make the default config static.
   workingDirectory: process.cwd(),
-  description: null,
+  description: undefined,
   allowExpressions: false,
 
   // FIXME(ofrobots): today we prioritize GAE_MODULE_NAME/GAE_MODULE_VERSION
   // over the user specified config. We should reverse that.
-  serviceContext: {service: null, version: null, minorVersion_: null},
+  serviceContext:
+      {service: undefined, version: undefined, minorVersion_: undefined},
 
-  appPathRelativeToRepository: null,
+  appPathRelativeToRepository: undefined,
   logLevel: 1,
   breakpointUpdateIntervalSec: 10,
   breakpointExpirationSec: 60 * 60 * 24,  // 24 hours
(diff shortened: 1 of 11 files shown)


src/agent/state.ts
```diff
@@ -29,9 +29,7 @@ import * as v8Types from '../types/v8-types';
 import * as apiTypes from '../types/api-types';
 import {DebugAgentConfig} from './config';
 
-// TODO: Determine if `ScopeType` should be named `scopeType`.
-// tslint:disable-next-line:variable-name
```
ofrobots: I would prefer to keep it capitalized as ScopeType. It is a foreign type, let's keep its original name.

test/test-debuglet.ts
```diff
@@ -1077,10 +1077,10 @@ describe('Debuglet', function() {
           'some project', 'id',
           // TODO: Verify that `null` for minorVersion_ should be used here
           //       or if minorVersion_ should be optional.
-          {service: 'some-service', version: 'production', minorVersion_: null},
+          {service: 'some-service', version: 'production', minorVersion_: undefined},
```
DominicKramer: I've removed `minorVersion_` and the TODO since they are no longer needed.

test/test-debuglet.ts
```diff
@@ -1077,10 +1077,10 @@ describe('Debuglet', function() {
...
           // TODO: Determine if these are the correct values that should be
           //       use here.
-          {}, null, null, false);
+          {}, false, undefined, undefined);
```
DominicKramer: The optional arguments can be omitted, and so I removed them.

test/test-debuglet.ts
```diff
@@ -1110,9 +1110,8 @@ describe('Debuglet', function() {
          //       fourth, or fifth parameters to be undefined.  Determine if
          //       this is correct.
          const debuggee = Debuglet.createDebuggee(
-             'a', 'b', undefined as any, undefined as any, undefined as any, 'Some Error Message',
              // TODO: Determine if this value for onGCP is correct.
-             false);
+             'a', 'b', undefined as any, undefined as any, false, undefined as any, 'Some Error Message');
```
DominicKramer: The TODO is not necessary.  I've removed it.  I've also replaced the `undefined as any` with the appropriate values without any casts.

test/test-debuglet.ts
```diff
@@ -1093,10 +1093,10 @@ describe('Debuglet', function() {
           Debuglet.createDebuggee('fancy-project', 'very-unique',
                                   // TODO: Verify that `null` for minorVersion_ should be used here
                                   //       or if minorVersion_ should be optional.
-                                  {service: 'default', version: 'yellow.5', minorVersion_: null},
+                                  {service: 'default', version: 'yellow.5', minorVersion_: undefined},
                                   // TODO: Determine if these are the correct values that should be
                                   //       use here.
-                                  {}, null, null, false);
+                                  {}, false, undefined, undefined);
```
DominicKramer: Done.

test/test-debuglet.ts
```diff
@@ -1093,10 +1093,10 @@ describe('Debuglet', function() {
           Debuglet.createDebuggee('fancy-project', 'very-unique',
                                   // TODO: Verify that `null` for minorVersion_ should be used here
                                   //       or if minorVersion_ should be optional.
-                                  {service: 'default', version: 'yellow.5', minorVersion_: null},
+                                  {service: 'default', version: 'yellow.5', minorVersion_: undefined},
```
DominicKramer: Done.

src/agent/state.ts
DominicKramer: I've reverted this change.

src/agent/config.ts
```diff
@@ -17,12 +17,12 @@
...
+  workingDirectory?: string;
 
   /**
    * A user specified way of identifying the service
    */
-  description: string|null;
+  description?: string;
 
   /**
    * Whether or not it is permitted to evaluate expressions.
```
matthewloring: Should `allowExpressions` and `serviceContext` be optional as well?

src/agent/debuglet.ts
```diff
@@ -108,22 +108,19 @@ const formatBreakpoints = function(
...
-  config_: DebugAgentConfig;
   private debug_: Debug;
   private v8debug_: V8DebugApi|null;
   private running_: boolean;
   private project_: string|null;
-  // TODO: Determine how to update the tests so that this can be private.
+  private debugletApi_: Controller;
+  private completedBreakpointMap_: {[key: string]: boolean};
+
+  // Exposed for testing
```
matthewloring: Should we leave the TODO on these? it doesn't seem like it was addressed. Or is it not possible to make these private?

DominicKramer: I don't know if we necessarily need them to be private at this point as long as we have it documented in the code that they are public for testing purposes.

src/agent/config.ts
DominicKramer: I think this would be best addressed in a follow-up PR where the config provided to the `start` method has all of its members as optional.

The default config, however, is of a different `ResolvedDebugAgentConfig` type where `allowExpressions`, `serviceContext`, and `workingDirectory` are not optional since they are resolved to specific values.

The config used throughout the code, then, is of the `ResolvedDebugAgentConfig` type where `allowExpressions`, `serviceContext`, and `workingDirectory` are not optional since their values are filled in from the default config if not specified in the `start` method.

This would simplify a lot of the type checks in the code, but I think should be in its own PR.

PR #325, opened 2017-08-23
Fix `.ts` files being copied to the build dir

DominicKramer:
Before this change, running `gulp` would incorrectly copy `.ts`
test files to the `build/test` directory.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -34,7 +34,9 @@ const tslintPath = path.join(__dirname, 'tslint.json');
 const outDir = 'build';
 const sources = ['src/**/*.ts'];
 const unitTests = ['test/**/*.ts'];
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -88,7 +88,7 @@ gulp.task('compile', () => {
 });
 
 gulp.task('test.system.copy', () => {
-  return gulp.src(['system-test/**/*'])
+  return gulp.src(['system-test/**/*.js'])
```
matthewloring: I think we may need to copy all extensions excluding `ts`. There are some `json` files for now I think and we may have other fixture type files in the future.

DominicKramer: Good catch.  Fixed.

PR #327, opened 2017-09-07
Change v8debugAPI clear to async.

gaofanmichael:
This PR aims to change the exisiting v8debug API clear function to async
in order to prepare for the migration debug api from v8 to inspector,
as inspector api is async.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -660,11 +660,14 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   removeBreakpoint_(breakpoint: Breakpoint): void {
+    const that = this;
     this.logger_.info('\tdeleted breakpoint', breakpoint.id);
     // TODO: Address the case when `breakpoint.id` is `undefined`.
     delete this.activeBreakpointMap_[breakpoint.id as string];
     if (this.v8debug_) {
-      this.v8debug_.clear(breakpoint);
+      this.v8debug_.clear(breakpoint, function(err) {
+        if (err) that.logger_.error(err);
+      });
     }
   }
 
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -660,13 +660,12 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   removeBreakpoint_(breakpoint: Breakpoint): void {
-    const that = this;
     this.logger_.info('\tdeleted breakpoint', breakpoint.id);
     // TODO: Address the case when `breakpoint.id` is `undefined`.
     delete this.activeBreakpointMap_[breakpoint.id as string];
     if (this.v8debug_) {
-      this.v8debug_.clear(breakpoint, function(err) {
-        if (err) that.logger_.error(err);
+      this.v8debug_.clear(breakpoint, (err) => {
+        if (err) this.logger_.error(err);
       });
     }
   }
--- a/src/agent/state.ts
+++ b/src/agent/state.ts
@@ -32,7 +32,7 @@ import {DebugAgentConfig} from './config';
 // TODO: Determine if `ScopeType` should be named `scopeType`.
 // tslint:disable-next-line:variable-name
 const ScopeType = vm.runInDebugContext('ScopeType');
-const assert = debugAssert(process.env.CLOUD_DEBUG_ASSERTIONS);
+const assert = debugAssert(!!process.env.CLOUD_DEBUG_ASSERTIONS);
 
 // Error message indices into the resolved variable table.
 const BUFFER_FULL_MESSAGE_INDEX = 0;
--- a/src/agent/v8debugapi.ts
+++ b/src/agent/v8debugapi.ts
@@ -37,6 +37,7 @@ const messages = {
       'A script matching the source file was not found loaded on the debuggee',
   SOURCE_FILE_AMBIGUOUS: 'Multiple files match the path specified',
   V8_BREAKPOINT_ERROR: 'Unable to set breakpoint in v8',
+  V8_BREAKPOINT_CLAER_ERROR: 'Unable to clear breakpoint in v8',
   SYNTAX_ERROR_IN_CONDITION: 'Syntax error in condition: ',
   ERROR_EVALUATING_CONDITION: 'Error evaluating condition: ',
   ERROR_COMPILING_CONDITION: 'Error compiling condition.',
@@ -54,7 +55,8 @@ const MODULE_WRAP_PREFIX_LENGTH = require('module').wrap('☃').indexOf('☃');
 
 export interface V8DebugApi {
   set: (breakpoint: apiTypes.Breakpoint, cb: (err: Error|null) => void) => void;
-  clear: (breakpoint: apiTypes.Breakpoint) => boolean;
+  clear:
+      (breakpoint: apiTypes.Breakpoint, cb: (err: Error|null) => void) => void;
   wait:
       (breakpoint: apiTypes.Breakpoint,
        callback: (err?: Error) => void) => void;
@@ -189,13 +191,18 @@ export function create(
       }
     },
 
-    clear: function(breakpoint: apiTypes.Breakpoint): boolean {
+    clear: function(
+        breakpoint: apiTypes.Breakpoint, cb: (err: Error|null) => void): void {
       if (typeof breakpoint.id === 'undefined') {
-        return false;
+        return setErrorStatusAndCallback(
+            cb, breakpoint, StatusMessage.BREAKPOINT_CONDITION,
+            messages.V8_BREAKPOINT_CLAER_ERROR);
       }
       const breakpointData = breakpoints[breakpoint.id];
       if (!breakpointData) {
-        return false;
+        return setErrorStatusAndCallback(
+            cb, breakpoint, StatusMessage.BREAKPOINT_CONDITION,
+            messages.V8_BREAKPOINT_CLAER_ERROR);
       }
       const v8bp = breakpointData.v8Breakpoint;
 
@@ -208,7 +215,9 @@ export function create(
         logger.info('deactivating v8 breakpoint listener');
         v8.setListener(null);
       }
-      return true;
+      return setImmediate(function() {
+        cb(null);
+      });
     },
 
     /**
--- a/src/agent/v8debugapi.ts
+++ b/src/agent/v8debugapi.ts
@@ -37,7 +37,7 @@ const messages = {
       'A script matching the source file was not found loaded on the debuggee',
   SOURCE_FILE_AMBIGUOUS: 'Multiple files match the path specified',
   V8_BREAKPOINT_ERROR: 'Unable to set breakpoint in v8',
-  V8_BREAKPOINT_CLAER_ERROR: 'Unable to clear breakpoint in v8',
+  V8_BREAKPOINT_CLEAR_ERROR: 'Unable to clear breakpoint in v8',
   SYNTAX_ERROR_IN_CONDITION: 'Syntax error in condition: ',
   ERROR_EVALUATING_CONDITION: 'Error evaluating condition: ',
   ERROR_COMPILING_CONDITION: 'Error compiling condition.',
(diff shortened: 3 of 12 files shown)


src/agent/debuglet.ts
```diff
@@ -660,11 +660,14 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   removeBreakpoint_(breakpoint: Breakpoint): void {
+    const that = this;
     this.logger_.info('\tdeleted breakpoint', breakpoint.id);
     // TODO: Address the case when `breakpoint.id` is `undefined`.
     delete this.activeBreakpointMap_[breakpoint.id as string];
     if (this.v8debug_) {
-      this.v8debug_.clear(breakpoint);
+      this.v8debug_.clear(breakpoint, function(err) {
```
DominicKramer: You should be able to use an arrow function here so that you can use `this` directly within the arrow function without having to use `const that = this;`.

test/test-v8debugapi.ts
```diff
@@ -409,8 +419,10 @@ describe('v8debugapi', function() {
           assert.ifError(err);
           api.wait(bp, function(err) {
             assert.ifError(err);
-            api.clear(bp);
-            done();
+            api.clear(bp, function(err) {
```
DominicKramer: Nit:  The  linter should have marked this `err` as shadowing the outer `err`, and required the inner `err` to have a different name (I usually use `err1` and `err2`).  I'm curious why the linter isn't flagging this.

test/test-v8debugapi.ts
```diff
@@ -476,8 +490,10 @@ describe('v8debugapi', function() {
         assert.ifError(err);
         api.wait(bp, function(err) {
           assert.ifError(err);
-          api.clear(bp);
-          done();
+          api.clear(bp, function(err) {
```
DominicKramer: Nit: The same name shadowing question as above.

test/test-v8debugapi.ts
```diff
@@ -589,9 +611,11 @@ describe('v8debugapi', function() {
           assert(((localsVal as any).status as any).description.format.match(
             'Locals and arguments are only displayed.*config.capture.maxExpandFrames=0'
             ));
-          api.clear(bp);
           config.capture.maxExpandFrames = oldCount;
```
DominicKramer: This line might need to be inside the callback to `api.clear`.  That is, the change to `config.capture.maxExpandFrames` could possibly happen before the call to `api.clear` runs in which case it is using the wrong value of `config.capture.maxExpandFrames`.

test/test-v8debugapi.ts
```diff
@@ -614,9 +638,11 @@ describe('v8debugapi', function() {
           assert.equal(topFrame['function'], 'foo');
           assert.equal(topFrame.locals[0].name, 'n');
           assert.equal(topFrame.locals[0].value, '2');
-          api.clear(bp);
           config.capture.maxFrames = oldMax;
```
DominicKramer: The same concern about moving this line inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -670,10 +696,12 @@ describe('v8debugapi', function() {
             return m.name === 'versions' && m.varTableIndex;
           } as any));
 
-          api.clear(bp);
           config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -723,9 +751,11 @@ describe('v8debugapi', function() {
             return resolved && (resolved.status as any).isError;
           }));
 
-          api.clear(bp);
           config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving this line inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -802,10 +834,12 @@ describe('v8debugapi', function() {
           const item = stringItems[0];
           assert(item.status.description.format.match(
             'Only first.*config.capture.maxStringLength=3.*of length 11.'));
-          api.clear(bp);
           config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -837,9 +871,11 @@ describe('v8debugapi', function() {
           assert((aVal as any).members[1].name.match(
             'Only first.*config.capture.maxProperties=1'));
 
-          api.clear(bp);
           config.capture.maxProperties = oldMax;
```
DominicKramer: The same concern about moving this line inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -870,9 +906,11 @@ describe('v8debugapi', function() {
           assert((bVal as any).members[1].name.match(
             'Only first.*config.capture.maxProperties=1'));
 
-          api.clear(bp);
           config.capture.maxProperties = oldMax;
```
DominicKramer: The same concern about moving this line inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -909,10 +947,12 @@ describe('v8debugapi', function() {
           // resulting in stringItems.length being 0.
           assert(stringItems.length === 1);
 
-          api.clear(bp);
           config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

src/agent/v8debugapi.ts
```diff
@@ -37,6 +37,7 @@ const messages = {
       'A script matching the source file was not found loaded on the debuggee',
   SOURCE_FILE_AMBIGUOUS: 'Multiple files match the path specified',
   V8_BREAKPOINT_ERROR: 'Unable to set breakpoint in v8',
+  V8_BREAKPOINT_CLAER_ERROR: 'Unable to clear breakpoint in v8',
```
ofrobots: nit: typo: CLAER vs CLEAR

test/test-v8debugapi.ts
```diff
@@ -946,10 +986,12 @@ describe('v8debugapi', function() {
             assert.equal(((fooVal as any).members as any).length, 4);
             assert.strictEqual((foo as any).status, undefined);
 
-            api.clear(bp);
             config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -982,10 +1024,12 @@ describe('v8debugapi', function() {
             assert.equal(((fooVal as any).members as any).length, 3);
             assert.strictEqual((foo as any).status, undefined);
 
-            api.clear(bp);
             config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -1019,10 +1063,12 @@ describe('v8debugapi', function() {
               'Max data size reached'));
             assert(((fooVal as any).status as any).isError);
 
-            api.clear(bp);
             config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -1343,11 +1407,15 @@ describe('v8debugapi', function() {
...
             assert.equal(api.numBreakpoints_(), 2);
-            api.clear(bp1);
-            assert.equal(api.numBreakpoints_(), 1);
-            api.clear(bp2);
-            assert.equal(api.numBreakpoints_(), 0);
-            done();
+            api.clear(bp1, function(err) {
+              assert.ifError(err);
+              assert.equal(api.numBreakpoints_(), 1);
+              api.clear(bp2, function(err) {
```
DominicKramer: Nit: The same concern of name shadowing as described above.

test/test-v8debugapi.ts
```diff
@@ -1097,10 +1145,12 @@ describe('v8debugapi', function() {
               'Max data size reached'));
             assert((bArray.status as any).isError);
 
-            api.clear(bp);
             config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

test/test-v8debugapi.ts
```diff
@@ -1056,10 +1102,12 @@ describe('v8debugapi', function() {
               'Max data size reached'));
             assert(((fooVal as any).status as any).isError);
 
-            api.clear(bp);
             config.capture.maxDataSize = oldMaxData;
```
DominicKramer: The same concern about moving these lines inside the callback of `api.clear` as described above.

src/agent/state.ts
```diff
@@ -32,7 +32,7 @@ import {DebugAgentConfig} from './config';
 // TODO: Determine if `ScopeType` should be named `scopeType`.
 // tslint:disable-next-line:variable-name
 const ScopeType = vm.runInDebugContext('ScopeType');
-const assert = debugAssert(process.env.CLOUD_DEBUG_ASSERTIONS);
+const assert = debugAssert(!!process.env.CLOUD_DEBUG_ASSERTIONS);
```
DominicKramer: Would the tests break without the addition of `!!` here?

src/agent/v8debugapi.ts
DominicKramer: Nit: There is a typo in the spelling of `CLEAR`.

test/test-debuglet.ts
```diff
@@ -289,7 +289,7 @@ describe('Debuglet', function() {
   });
 
   describe('setup', function() {
-    before(function() { oldGP = process.env.GCLOUD_PROJECT; });
+    before(function() { oldGP = String(process.env.GCLOUD_PROJECT); });
```
DominicKramer: If `process.env.GCLOUD_PROJECT` is `undefined`, then it looks like `oldGP` will be set to the string `'undefined'` which could cause reseting the value of `process.env.GCLOUD_PROJECT` later to not work as expected.

test/test-debuglet.ts
```diff
@@ -436,7 +436,7 @@ describe('Debuglet', function() {
 
       it('should respect GCLOUD_DEBUG_LOGLEVEL', function(done) {
         process.env.GCLOUD_PROJECT = '11020304f2934';
-        process.env.GCLOUD_DEBUG_LOGLEVEL = 3;
+        process.env.GCLOUD_DEBUG_LOGLEVEL = '3';
```
DominicKramer: Nit: This shouldn't be needed because `process.env` is special in that when you set an attribute on `process.env` it will be automatically cast to a string.

src/agent/v8debugapi.ts
gaofanmichael: resolve.

src/agent/v8debugapi.ts
```diff
@@ -208,7 +215,9 @@ export function create(
         logger.info('deactivating v8 breakpoint listener');
         v8.setListener(null);
       }
-      return true;
+      return setImmediate(function() {
```
matthewloring: No need to change this but for future reference this can be written: `setImmediate(cb, null)`.

test/test-debuglet.ts
matthewloring: We don't yet have style best practices for typescript but I think using Typescript casts (`process.env.GCLOUD_PROJECT as string`) instead of Javascript coersions (`String(process.env.GCLOUD_PROJECT)`) is preferable just so that the actual behavior at runtime is not changed (especially to avoid bugs such as the one @DominicKramer described).

src/agent/state.ts
gaofanmichael: As debugAssert takes boolean and process.env.CLOUD_DEBUG_ASSERTIONS is a null or string. Converting to boolean value is needed.

test/test-debuglet.ts
gaofanmichael: Redefine oldGP variable to be string|undefined in test-debuglet.ts

test/test-debuglet.ts
gaofanmichael: Without manually converting, it gives error as type '3' is not assigned to string|null

test/test-v8debugapi.ts
gaofanmichael: as in the whole file, such as where api.set and then api.wait, the code uses only 'err', it looks like linter would't complain for this. I will not change for now and we can follow up on this issue.

src/agent/state.ts
kjin: @DominicKramer `@types/node` changed the type of `process.env` from `any` to `{ [p:string]: string|undefined }` from v7 to v8.

PR #330, opened 2017-09-18
Adjust api breakpoint if resolved in another line

gaofanmichael:
Change api type breakpoint location line number if the breakpoint was
originally set to a line with only comments. The new line that
breakpoint set to will be actual paused line in code.

### DIFF
--- a/src/agent/state.ts
+++ b/src/agent/state.ts
@@ -169,7 +169,6 @@ class StateResolver {
     // evaluated expressions can be evaluated as much as possible within
     // the max data size limits
     const frames = that.resolveFrames_();
-
     // Now resolve the variables
     let index = this.messageTable_.length;  // skip the sentinel values
     const noLimit = that.config_.capture.maxDataSize === 0;
@@ -190,8 +189,9 @@ class StateResolver {
     if (index < that.rawVariableTable_.length) {
       that.trimVariableTable_(index, frames);
     }
-
     return {
+      location: {line: this.state_.frame(0).sourceLine()} as
+          apiTypes.SourceLocation,
       stackFrames: frames,
       variableTable: that.resolvedVariableTable_,
       evaluatedExpressions: that.evaluatedExpressions_
--- a/src/agent/state.ts
+++ b/src/agent/state.ts
@@ -190,6 +190,8 @@ class StateResolver {
       that.trimVariableTable_(index, frames);
     }
     return {
+      // TODO (fgao): Add path attribute to avoid explicit cast to
+      // apiTypes.SourceLocation once breakpoint is passed in this class.
       location: {line: this.state_.frame(0).sourceLine()} as
           apiTypes.SourceLocation,
       stackFrames: frames,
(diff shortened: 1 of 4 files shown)


src/agent/state.ts
```diff
@@ -190,8 +189,9 @@ class StateResolver {
     if (index < that.rawVariableTable_.length) {
       that.trimVariableTable_(index, frames);
     }
-
     return {
+      location: {line: this.state_.frame(0).sourceLine()} as
```
DominicKramer: Is it possible to add the `path` attribute so that an explicit cast to `apiType.SourceLocation` is not needed?

src/agent/v8debugapi.ts
```diff
@@ -612,6 +612,11 @@ export function create(
       // TODO: Address the case where `breakpoint.expression` is `undefined`.
       const captured = state.capture(
           execState, breakpoint.expressions as string[], config, v8);
+      if (breakpoint.location !== undefined &&
```
DominicKramer: Nit: It looks like `breakpoint.location` and `captured.location` can be checked directly (i.e. `if(breakpoint.location && ...) { ... }`).

Why is it necessary to check that `breakpoint.location.line !== undefined`?

test/test-v8debugapi.ts
```diff
@@ -500,6 +500,29 @@ describe('v8debugapi', function() {
 
     });
 
+    it('should resolve correct breakpoint line number', function(done) {
+      // clone a clean breakpointInFoo
+      // TODO: Have this actually implement Breakpoint
```
DominicKramer: Is it possible to address these two TODOs in this PR?

test/test-v8debugapi.ts
```diff
@@ -500,6 +500,29 @@ describe('v8debugapi', function() {
 
     });
 
+    it('should resolve correct breakpoint line number', function(done) {
```
matthewloring: Could the test name be updated to mention why the line number may have been incorrect?

src/agent/v8debugapi.ts
gaofanmichael: Not necessary. removed.

test/test-v8debugapi.ts
gaofanmichael: copied from other part. not TODO needed here. I will remove these TODOs.

src/agent/state.ts
gaofanmichael: It is possible. However, I will need to pass the breakpoint to state.ts and use the path in original breakpoint. I will add a TODO here once we have breakpoint passed in from another PR I am working on.

src/agent/v8debugapi.ts
```diff
@@ -612,6 +612,8 @@ export function create(
       // TODO: Address the case where `breakpoint.expression` is `undefined`.
       const captured = state.capture(
           execState, breakpoint.expressions as string[], config, v8);
+      if (breakpoint.location && captured.location && captured.location.line)
```
matthewloring: Can we move this +1 up to the place where `captured.location.line` is set?

kjin: nit: use {} even for one-line inner statements, it helps with code readability.

PR #332, opened 2017-09-21
Add dependency on extend@3

kjin:
We use `extend` but don't depend on it. This change fixes that.


kjin:
@matthewloring Yep! Agreed

PR #333, opened 2017-09-26
Fix system test break

gaofanmichael:
Fixed the e2e test when logpoint and breakpoint was set in one place.

### DIFF
--- a/src/agent/inspectordebugapi.ts
+++ b/src/agent/inspectordebugapi.ts
@@ -33,7 +33,7 @@ import {V8Inspector} from './v8inspector';
 
 export class BreakpointData {
   constructor(
-      public id: string, public active: boolean,
+      public id: inspector.Debugger.BreakpointId, public active: boolean,
       public apiBreakpoint: apiTypes.Breakpoint,
       public parsedCondition: estree.Node, public locationStr: string,
       public compile: null|((src: string) => string)) {}
@@ -46,12 +46,17 @@ export class InspectorDebugApi implements debugapi.DebugApi {
   breakpoints: {[id: string]: BreakpointData} = {};
   sourcemapper: SourceMapper;
   session: inspector.Session;
+  // TODO: listeners, scrpitmapper, location mapper and breakpointmapper can use
+  // Map in the future after resolving Map.prototype.get(key) returns V or
+  // undefined.
   listeners: {[id: string]: utils.Listener} = {};
+  // scriptmapper maps scriptId to actual script path.
   scriptmapper: {[id: string]: any} = {};
-  // locationmapper maps location string to a list of breakpoint.id.
-  locationmapper: {[id: string]: string[]} = {};
-  // breakpointmapper maps inspector breakpoint id to a list of breakpoint.id.
-  breakpointmapper: {[id: string]: string[]} = {};
+  // locationmapper maps location string to a list of stackdriver breakpoint id.
+  locationmapper: {[id: string]: apiTypes.BreakpointId[]} = {};
+  // breakpointmapper maps v8/inspector breakpoint id to a list of
+  // stackdriver breakpoint id.
+  breakpointmapper: {[id: string]: apiTypes.BreakpointId[]} = {};
   numBreakpoints = 0;
   v8Inspector: V8Inspector;
   constructor(
@@ -131,22 +136,24 @@ export class InspectorDebugApi implements debugapi.DebugApi {
           utils.messages.V8_BREAKPOINT_CLEAR_ERROR);
     }
     let locationStr = breakpointData.locationStr;
-    const bpId = breakpointData.id;
+    const v8BreakpointId = breakpointData.id;
 
-    // delete breakpoint from locationmapper and breakpointmapper.
+    // delete current breakpoint from locationmapper and breakpointmapper.
     this.deleteBreakpointInArray(
         this.locationmapper[locationStr], breakpoint.id);
     if (this.locationmapper[locationStr].length === 0) {
       delete this.locationmapper[locationStr];
     }
-    this.deleteBreakpointInArray(this.breakpointmapper[bpId], breakpoint.id);
-    if (this.breakpointmapper[bpId].length === 0) {
-      delete this.breakpointmapper[bpId];
+    this.deleteBreakpointInArray(
+        this.breakpointmapper[v8BreakpointId], breakpoint.id);
+    if (this.breakpointmapper[v8BreakpointId].length === 0) {
+      delete this.breakpointmapper[v8BreakpointId];
     }
 
     let result: {error?: Error} = {};
     if (!this.breakpointmapper[breakpointData.id]) {
-      // No breakpoint for this inspector brekapoint.
+      // When breakpointmapper does not countain current v8/inspector breakpoint
+      // id, we should remove this breakpoint from v8.
       result = this.v8Inspector.removeBreakpoint(breakpointData.id);
     }
     delete this.breakpoints[breakpoint.id];
@@ -227,6 +234,7 @@ export class InspectorDebugApi implements debugapi.DebugApi {
   }
 
   numBreakpoints_(): number {
+    // Tracks the number of stackdriver breakpoints.
     return Object.keys(this.breakpoints).length;
   }
 
@@ -344,9 +352,13 @@ export class InspectorDebugApi implements debugapi.DebugApi {
     }
     let condition;
     if (breakpoint.condition) condition = breakpoint.condition;
-    let locationStr = JSON.stringify(breakpoint.location);
 
-    let bpId;  // inspector breakpoint id
+    // location Str will be a JSON string of Stackdriver breakpoint location.
+    // It will be used as key at locationmapper to ensure there will be no
+    // duplicate breakpoints at the same location.
+    let locationStr = JSON.stringify(breakpoint.location);
+    let v8BreakpointId;  // v8/inspector breakpoint id
+
     if (!this.locationmapper[locationStr]) {
       // The first time when a breakpoint was set to this location.
       let res = this.v8Inspector.setBreakpointByUrl(
@@ -356,18 +368,23 @@ export class InspectorDebugApi implements debugapi.DebugApi {
             cb, breakpoint, StatusMessage.BREAKPOINT_SOURCE_LOCATION,
             utils.messages.V8_BREAKPOINT_ERROR);
       }
-      bpId = res.response.breakpointId;
+      v8BreakpointId = res.response.breakpointId;
       this.locationmapper[locationStr] = [];
-      this.breakpointmapper[bpId] = [];
+      this.breakpointmapper[v8BreakpointId] = [];
     } else {
-      // Breakpoint found on this location. Acquire the breakpoint id.
-      bpId = this.breakpoints[this.locationmapper[locationStr][0]].id;
+      // Breakpoint found at this location. Acquire the v8/inspector breakpoint
+      // id.
+      v8BreakpointId = this.breakpoints[this.locationmapper[locationStr][0]].id;
     }
 
+    // Adding current stackdriver breakpoint id to location mapper and
+    // breakpoint mapper.
     this.locationmapper[locationStr].push(breakpoint.id as string);
-    this.breakpointmapper[bpId].push(breakpoint.id as string);
+    this.breakpointmapper[v8BreakpointId].push(breakpoint.id as string);
+
     this.breakpoints[breakpoint.id as string] = new BreakpointData(
-        bpId, true, breakpoint, ast as estree.Program, locationStr, compile);
+        v8BreakpointId, true, breakpoint, ast as estree.Program, locationStr,
+        compile);
 
     this.numBreakpoints++;
     setImmediate(function() {
@@ -441,9 +458,9 @@ export class InspectorDebugApi implements debugapi.DebugApi {
         const evaluatedExpressions = breakpoint.expressions.map(function(exp) {
           const result = state.evaluate(exp, frame, that.v8Inspector, true);
           // TODO: Address the case where `result.mirror` is `undefined`.
-          if (result.error)
+          if (result.error) {
             return result.error;
-          else {
+          } else {
             return (result.object as inspector.Runtime.RemoteObject).value;
           }
         });
@@ -469,8 +486,8 @@ export class InspectorDebugApi implements debugapi.DebugApi {
                                      inspector.Debugger.PausedEventDataType) {
     try {
       if (!params.hitBreakpoints) return;
-      const bpId: string = params.hitBreakpoints[0];
-      this.breakpointmapper[bpId].forEach((id: string) => {
+      const v8BreakpointId: string = params.hitBreakpoints[0];
+      this.breakpointmapper[v8BreakpointId].forEach((id: string) => {
         if (this.listeners[id].enabled) {
           this.logger.info('>>>breakpoint hit<<< number: ' + id);
           this.listeners[id].listener(params.callFrames);
--- a/src/agent/inspectordebugapi.ts
+++ b/src/agent/inspectordebugapi.ts
@@ -51,12 +51,12 @@ export class InspectorDebugApi implements debugapi.DebugApi {
   // undefined.
   listeners: {[id: string]: utils.Listener} = {};
   // scriptmapper maps scriptId to actual script path.
-  scriptmapper: {[id: string]: any} = {};
+  scriptMapper: {[id: string]: any} = {};
   // locationmapper maps location string to a list of stackdriver breakpoint id.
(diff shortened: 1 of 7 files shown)


src/agent/inspectordebugapi.ts
```diff
@@ -35,7 +35,7 @@ export class BreakpointData {
   constructor(
       public id: string, public active: boolean,
       public apiBreakpoint: apiTypes.Breakpoint,
-      public parsedCondition: estree.Node,
+      public parsedCondition: estree.Node, public locationStr: string,
```
DominicKramer: nit: The `public locationStr: string` should be on the next line.  If the formatter put it on the line that it is on, we will need to investigate how to configure the formatter.

src/agent/inspectordebugapi.ts
```diff
@@ -46,9 +46,12 @@ export class InspectorDebugApi implements debugapi.DebugApi {
   breakpoints: {[id: string]: BreakpointData} = {};
   sourcemapper: SourceMapper;
   session: inspector.Session;
-  scriptmapper: {[id: string]: any} = {};
-
   listeners: {[id: string]: utils.Listener} = {};
+  scriptmapper: {[id: string]: any} = {};
```
DominicKramer: For `scriptmapper`, `locationmapper`, and `breakpointmapper`, could a `Map<>` be used instead of an Object.

src/agent/inspectordebugapi.ts
```diff
@@ -143,25 +162,26 @@ export class InspectorDebugApi implements debugapi.DebugApi {
 
   wait(breakpoint: apiTypes.Breakpoint, callback: (err?: Error) => void): void {
     // TODO: Address the case whree `breakpoint.id` is `null`.
-    const bp = this.breakpoints[breakpoint.id as string];
     const listener =
         this.onBreakpointHit.bind(this, breakpoint, (err: Error) => {
-          this.listeners[bp.id].enabled = false;
+          this.listeners[breakpoint.id as string].enabled = false;
```
DominicKramer: Why is this cast to a `string` needed?

src/agent/inspectordebugapi.ts
```diff
@@ -394,11 +439,13 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       } else {
         const frame = callFrames[0];
         const evaluatedExpressions = breakpoint.expressions.map(function(exp) {
-          const result = state.evaluate(exp, frame, that.v8Inspector);
+          const result = state.evaluate(exp, frame, that.v8Inspector, true);
           // TODO: Address the case where `result.mirror` is `undefined`.
-          return result.error ?
-              result.error :
-              (result.object as inspector.Runtime.RemoteObject).value();
+          if (result.error)
```
DominicKramer: To reduce the chances of introducing subtle bugs, `if` statements should use braces, even if the body contains a single statement.

src/agent/v8inspector.ts
```diff
@@ -55,10 +55,9 @@ export class V8Inspector {
   }
 
   evaluateOnCallFrame(
-      callFrameId: string, expression: string, objectGroup?: string,
-      includeCommandLineAPI?: boolean, silent?: boolean,
-      returnByValue?: boolean, generatePreview?: boolean,
-      throwOnSideEffect?: boolean) {
+      callFrameId: string, expression: string, returnByValue?: boolean,
```
DominicKramer: Why did the order of these parameters change?

src/agent/inspectordebugapi.ts
```diff
@@ -127,11 +130,27 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
-    const id = breakpoint.id;
-    const result = this.v8Inspector.removeBreakpoint(breakpointData.id);
-    const breakpointId = this.breakpoints[id].id;
-    delete this.breakpoints[id];
-    delete this.listeners[breakpointId];
+    let locationStr = breakpointData.locationStr;
+    const bpId = breakpointData.id;
+
+    // delete breakpoint from locationmapper and breakpointmapper.
+    this.deleteBreakpointInArray(
```
DominicKramer: It looks like it would be advantageous to have `locationmapper` and `breakpointmapper` map keys to a `Map<string>` instead of a `string[]` to support deletion here.

src/agent/inspectordebugapi.ts
gaofanmichael: listener was used v8 breakpoint ID and now uses stackdriver breakpoint id. type from string to string|null.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/agent/v8inspector.ts
gaofanmichael: Since I will only use callframeId, expression and returnByValue, i will move these three together so that when I call the function, I can just pass three variables.  the remaining parameters is matched with protocal (which will not likely to use for now)

src/agent/inspectordebugapi.ts
gaofanmichael: gulp format resolved to same line.

src/agent/inspectordebugapi.ts
gaofanmichael: Map.prototype.get returns value or undefined. This makes whereever we want to use the code for get be casting to the actual value type, and makes it complex. I added a TODO here and once we have a wrapper of Map class to implement Map.prototype.get returns the actually value (or throw exception), and we can replace all map object in the code to be Map.

gaofanmichael:
I signed it!

src/agent/inspectordebugapi.ts
```diff
@@ -46,9 +46,17 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
   sourcemapper: SourceMapper;
   session: inspector.Session;
-  scriptmapper: {[id: string]: any} = {};
-
+  // TODO: listeners, scrpitmapper, location mapper and breakpointmapper can use
+  // Map in the future after resolving Map.prototype.get(key) returns V or
+  // undefined.
   listeners: {[id: string]: utils.Listener} = {};
+  // scriptmapper maps scriptId to actual script path.
+  scriptmapper: {[id: string]: any} = {};
```
kjin: Use camelCase, for scriptmapper, locationmapper, and breakpointmapper

ofrobots:
Hmm.. what happened with the CLA here? Do you somehow have a commit with a different address set in your git config?

src/agent/inspectordebugapi.ts
```diff
@@ -181,36 +208,47 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
   numBreakpoints_(): number {
+    // Tracks the number of stackdriver breakpoints.
     return Object.keys(this.breakpoints).length;
   }
 
   numListeners_(): number {
     return Object.keys(this.listeners).length;
   }
 
+  private deleteBreakpointInArray(breakpointArray: string[], bp: string) {
```
kjin: Make this `static`.

gaofanmichael:
I have not modified my git config. I added my corp email address to CLA list and it worked. I have no ideas why googlebot checked cla:no only this time.

src/agent/inspectordebugapi.ts
```diff
@@ -312,37 +350,65 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
     });  // success.
   }
 
+  private setAndStoreBreakpoint(
+      breakpoint: apiTypes.Breakpoint, line: number, column: number,
+      matchingScript: string):
+      {v8BreakpointId: inspector.Debugger.BreakpointId,
+       locationStr: string}|null {
+    let condition;
+    if (breakpoint.condition) condition = breakpoint.condition;
```
kjin: Can you use brackets here?

src/agent/inspectordebugapi.ts
```diff
@@ -312,37 +350,65 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
+    // location Str will be a JSON string of Stackdriver breakpoint location.
+    // It will be used as key at locationmapper to ensure there will be no
+    // duplicate breakpoints at the same location.
+    let locationStr = JSON.stringify(breakpoint.location);
+    let v8BreakpointId;  // v8/inspector breakpoint id
+    if (!this.locationmapper[locationStr]) {
+      // The first time when a breakpoint was set to this location.
+      let res = this.v8Inspector.setBreakpointByUrl(
+          line - 1, matchingScript, undefined, column - 1, condition);
```
kjin: I'm surprised the compiler doesn't complain that of use before assignment for `condition`. I think you can directly set the last argument to `breakpoint.condition` and remove lines 377-378.

src/agent/inspectordebugapi.ts
```diff
@@ -312,37 +350,65 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
+      breakpoint: apiTypes.Breakpoint, line: number, column: number,
+      matchingScript: string):
+      {v8BreakpointId: inspector.Debugger.BreakpointId,
+       locationStr: string}|null {
+    let condition;
+    if (breakpoint.condition) condition = breakpoint.condition;
+    // location Str will be a JSON string of Stackdriver breakpoint location.
+    // It will be used as key at locationmapper to ensure there will be no
+    // duplicate breakpoints at the same location.
+    let locationStr = JSON.stringify(breakpoint.location);
```
kjin: Prefer `const`

src/agent/inspectordebugapi.ts
```diff
@@ -312,37 +350,65 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
+      // id.
+      v8BreakpointId = this.breakpoints[this.locationmapper[locationStr][0]].id;
+    }
+
+    // Adding current stackdriver breakpoint id to location mapper and
+    // breakpoint mapper.
+    this.locationmapper[locationStr].push(breakpoint.id as string);
+    this.breakpointmapper[v8BreakpointId].push(breakpoint.id as string);
+
+    return {v8BreakpointId: v8BreakpointId, locationStr: locationStr};
```
kjin: Can be shorthanded as `{ v8BreakpointId, locationStr }`.

src/agent/inspectordebugapi.ts
```diff
@@ -312,37 +350,65 @@ export class InspectorDebugApi implements debugapi.DebugApi {
...
+  }
+
   private onBreakpointHit(
       breakpoint: apiTypes.Breakpoint, callback: (err: Error|null) => void,
       callFrames: inspector.Debugger.CallFrame[]): void {
-    // TODO: Address the situation where `breakpoint.id` is `null`.
-    const bp = this.breakpoints[breakpoint.id as string];
-    if (!bp.active) {
```
kjin: Where did this block go?

src/agent/inspectordebugapi.ts
```diff
@@ -394,11 +460,13 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       } else {
         const frame = callFrames[0];
         const evaluatedExpressions = breakpoint.expressions.map(function(exp) {
-          const result = state.evaluate(exp, frame, that.v8Inspector);
+          const result = state.evaluate(exp, frame, that.v8Inspector, true);
```
kjin: Can you add a comment as to why we are passing in `true` here?

src/agent/v8inspector.ts
kjin: If you have more than 3 arguments to a function, it might be worthwhile to accept a single `options` object instead, and take advantage of required and optional fields in TypeScript. This way you don't have to re-order arguments to enhance code readability.

src/types/api-types.ts
```diff
@@ -78,7 +78,7 @@ export interface Breakpoint {
   variableTable: Array<Variable|null>;
   // TODO: The `controller.ts` file assumes `id` is not null or undefined.
   //       Verify it it should be optional.
-  id?: string;
+  id?: BreakpointId;
```
kjin: You can actually make this non-optional and remove the TODO. Looking at the [docs](https://cloud.google.com/debugger/api/reference/rest/v2/debugger.debuggees.breakpoints) each of the endpoints that yield `Breakpoint` objects, every description states that `id` is guaranteed to be set.

This way, you can also get rid of the `as string` casts elsewhere.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/agent/inspectordebugapi.ts
gaofanmichael: Removed.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/agent/inspectordebugapi.ts
gaofanmichael: Per our discussion, inspector protocol cannot check the v8 breakpoint active status. And theoretically breakpoint will not be inactive as we never disable them. I remove this part of code and remove the active field in BreakpointData.

src/agent/inspectordebugapi.ts
gaofanmichael: Done.

src/types/api-types.ts
gaofanmichael: Done.

src/agent/inspectordebugapi.ts
gaofanmichael: Moved to utils.

src/agent/utils.ts
```diff
@@ -178,3 +178,10 @@ export function getBreakpointCompiler(breakpoint: apiTypes.Breakpoint):
       return null;
   }
 }
+
+export function removeElementInArray(array: string[], element: string) {
```
kjin: nit: `removeFirstOccurrenceInArray<T>(array: T[], element: T): void`

src/agent/inspectordebugapi.ts
```diff
@@ -33,9 +33,9 @@ import {V8Inspector} from './v8inspector';
```
ofrobots: Not part of this change, so this can be fixed in a follow-on. It would be good to use snake-case for file names.

src/agent/v8debugapi.ts
```diff
@@ -187,7 +187,7 @@ export class V8DebugApi implements debugapi.DebugApi {
       shouldStop: () => boolean): void {
     const that = this;
     // TODO: Address the case whree `breakpoint.id` is `null`.
```
ofrobots: nit: this TODO can be removed now.

src/agent/v8debugapi.ts
```diff
@@ -338,7 +338,7 @@ export class V8DebugApi implements debugapi.DebugApi {
       this.v8.setListener(this.handleDebugEvents);
     }
     // TODO: Address the case whree `breakpoint.id` is `null`.
```
ofrobots: ditto

src/agent/v8debugapi.ts
```diff
@@ -360,7 +360,7 @@ export class V8DebugApi implements debugapi.DebugApi {
       breakpoint: apiTypes.Breakpoint, callback: (err: Error|null) => void,
       execState: v8Types.ExecutionState): void {
     // TODO: Address the situation where `breakpoint.id` is `null`.
-    const v8bp = this.breakpoints[breakpoint.id as string].v8Breakpoint;
+    const v8bp = this.breakpoints[breakpoint.id].v8Breakpoint;
```
ofrobots: ditto

src/agent/v8debugapi.ts
```diff
@@ -422,15 +422,14 @@ export class V8DebugApi implements debugapi.DebugApi {
       execState: v8Types.ExecutionState): void {
     const expressionErrors: Array<apiTypes.Variable|null> = [];
     // TODO: Address the case where `breakpoint.id` is `null`.
```
ofrobots: ditto

src/agent/v8debugapi.ts
gaofanmichael: Done.

src/agent/v8debugapi.ts
gaofanmichael: Done.

src/agent/v8debugapi.ts
gaofanmichael: Done.

src/agent/v8debugapi.ts
gaofanmichael: Done.

PR #335, opened 2017-10-02
Implement side-effect-free expression evaluation

gaofanmichael:
Use flag throwOnSideEffect from debugger.onEvaluateOnCallframe to implement side-effect-free.


ofrobots:
The change itself looks reasonable, but how do we ensure that we are still providing _useful_ representation of objects after this change?

For example it is valid (as far as our tests are concerned) to evaluate a `request` object as:

```
request: {
  // no properties available as all of them were deemed to have side effects
}
```

This would degrade the user experience.

I would suggest adding some tests that ensure that some common data structures are still evaluated with same/similar fidelity as before this change.

DominicKramer:
@ofrobots The tests you suggested have been implemented in #425.  Do you think this PR looks good to land now?

test/test-expression-side-effect.ts
```diff
@@ -0,0 +1,148 @@
...
+import {Common, LoggerOptions} from '../src/types/common';
+import * as assert from 'assert';
+import {StatusMessage} from '../src/client/stackdriver/status-message';
+import * as utils from '../src/agent/util/utils';
+
+const code = require('./test-expression-side-effect-code.js');
+
+const common: Common = require('@google-cloud/common');
+
+const nodeAtLeast8withInspector = utils.satisfies(process.version, '>=8') && process.env.GCLOUD_USE_INSPECTOR;
```
ofrobots: Does this pass `gts check` for formatting?

test/test-expression-side-effect.ts
```diff
@@ -0,0 +1,148 @@
...
+const nodeAtLeast8OrInspector = utils.satisfies(process.version, '>=8') || process.env.GCLOUD_USE_INSPECTOR;
+const nodeAtLeast9withInspector = utils.satisfies(process.version, '>=9.1.0') && process.env.GCLOUD_USE_INSPECTOR;
```
ofrobots: I am not sure if all these combinations still make sense. My expectation is that `GCLOUD_USE_INSPECTOR` should be the default for Node 10+. I would prefer that `GCLOUD_USE_INSPECTOR` go away. That is not this PR though, but there is complexity in this one because of that env var.

Can we remove that env. var. in a separate PR and rebase this on top?

test/test-v8debugapi.ts
```diff
@@ -1559,55 +1559,50 @@ describe('v8debugapi', () => {
...
-                       expr!.status!.description.format,
-                       'Error Compiling Expression');
-                 } else {
-                   assert(expr!.status!.description.format.match(
-                       'Unexpected token'));
-                 }
-               }
-             }
+      function(done) {
+        // TODO: Have this actually implement Breakpoint
```
ofrobots: nit: TODOs should have owners.

test/test-expression-side-effect.ts
DominicKramer: Sure

test/test-v8debugapi.ts
DominicKramer: Done

test/test-expression-side-effect.ts
```diff
@@ -0,0 +1,154 @@
...
+import * as debugapi from '../src/agent/v8/debugapi';
+import {StatusMessage} from '../src/client/stackdriver/status-message';
+import {Common, LoggerOptions} from '../src/types/common';
+import * as stackdriver from '../src/types/stackdriver';
+
+const code = require('./test-expression-side-effect-code.js');
+
+const common: Common = require('@google-cloud/common');
+
+const itWithInspector = utils.satisfies(process.version, '>=10') ? it : it.skip;
```
ofrobots: nit: add a comment that we use inspector as the default on 10+, hence the skip on earlier versions.

PR #340, opened 2017-10-11
Elaborate on async stack trace warning

gaofanmichael:
Add a log message address the 'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE' warning.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -30,6 +30,7 @@ import * as _ from 'lodash';
 import * as path from 'path';
 import * as semver from 'semver';
 import * as util from 'util';
+import * as utils from './util/utils';
 import * as http from 'http';
 
 import {Controller} from './controller';
@@ -205,17 +206,14 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   async start(): Promise<void> {
+    const that = this;
     process.on('warning', (warning) => {
       if ((warning as any).code ===
           'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE') {
-        console.log(
-            'The Stackdriver debug agent does not use Inspector async stack ' +
-            'traces. The INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE warning ' +
-            'does not impact Stackdriver debug agent and can be ignored.');
+        that.logger_.info(utils.messages.ASYNC_TRACES_WARNING);
       }
     });
 
-    const that = this;
     const stat = promisify(fs.stat);
 
     try {
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -205,6 +205,14 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   async start(): Promise<void> {
+    process.on('warning', (warning) => {
+      if ((warning as any).code ===
+          'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE') {
+        console.log(
+            'The current debug agent does not use Inspector async stack ' +
+            'traces. The above warning will not affect the debug agent.');
+      }
+    });
     const that = this;
     const stat = promisify(fs.stat);
 
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -209,8 +209,9 @@ export class Debuglet extends EventEmitter {
       if ((warning as any).code ===
           'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE') {
         console.log(
-            'The current debug agent does not use Inspector async stack ' +
-            'traces. The above warning will not affect the debug agent.');
+            'The Stackdriver debug agent does not use Inspector async stack ' +
+            'traces. The INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE warning ' +
+            'does not impact Stackdriver debug agent and can be ignored.');
       }
     });
 
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -213,6 +213,7 @@ export class Debuglet extends EventEmitter {
             'traces. The above warning will not affect the debug agent.');
       }
     });
+
     const that = this;
     const stat = promisify(fs.stat);
 
--- a/src/agent/util/utils.ts
+++ b/src/agent/util/utils.ts
@@ -24,7 +24,10 @@ export const messages = {
   CAPTURE_BREAKPOINT_DATA: 'Error trying to capture snapshot data: ',
   INVALID_LINE_NUMBER: 'Invalid snapshot position: ',
   COULD_NOT_FIND_OUTPUT_FILE:
-      'Could not determine the output file associated with the transpiled input file'
+      'Could not determine the output file associated with the transpiled input file',
+  ASYNC_TRACES_WARNING: 'The Stackdriver debug agent does not use Inspector ' +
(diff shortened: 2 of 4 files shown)


src/agent/debuglet.ts
```diff
@@ -205,6 +205,15 @@ export class Debuglet extends EventEmitter {
    * @private
    */
   async start(): Promise<void> {
+    process.on('warning', (warning) => {
+      if ((warning as any).code ===
+          'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE') {
+        console.log(
+            'The current debug agent does not use Inspector async stack ' +
+            'traces. The above warning will not affect the debug agent.');
```
DominicKramer: Do we know for sure that the warning message will be printed before the `console.log` message?  If not, the use of `The above warning` wouldn't be correct.

test/test-debuglet.ts
```diff
@@ -315,6 +316,49 @@ describe('Debuglet', function() {
       assert.deepEqual(mergedConfig, compareConfig);
     });
 
+    it('should elaborate on inspector warning on 32 bit', function(done) {
+      let logText = '';
+      const oldLog = console.log;
+      console.log = function(s: string) {
```
DominicKramer: I would suggest setting `console.log` in `beforeEach` and restoring `console.log` if `afterEach`.  Otherwise, if the test fails, the code for restoring `console.log` won't get run, which could influence any tests that follow.

test/test-debuglet.ts
```diff
@@ -315,6 +316,49 @@ describe('Debuglet', function() {
       assert.deepEqual(mergedConfig, compareConfig);
     });
 
+    it('should elaborate on inspector warning on 32 bit', function(done) {
```
DominicKramer: Since the test also tests that the message doesn't appear on 64 bit machines, I think it would be good to add ` but not on 64 bit` to the test name.

test/test-debuglet.ts
```diff
@@ -315,6 +316,49 @@ describe('Debuglet', function() {
...
+
+      debuglet.once('registered', function(id: string) {
+        assert.equal(id, DEBUGGEE_ID);
+        // TODO: Handle the case where debuglet.debuggee is undefined
+        assert.equal((debuglet.debuggee_ as Debuggee).project, projectId);
+        console.log = oldLog;
+        const arch = process.arch;
+        const nodeVersion = /v(\d+\.\d+\.\d+)/.exec(process.version);
+        if (!nodeVersion || nodeVersion.length < 2) {
+          (assert as any).fail();
```
DominicKramer: Why is the cast to `any` needed here?

test/test-debuglet.ts
```diff
@@ -315,6 +316,49 @@ describe('Debuglet', function() {
...
+        } else if (semver.satisfies(nodeVersion[1], '>=8') &&
+            (arch === 'ia32' || arch === 'x86')) {
+          assert(logText.includes(
+              'The current debug agent does not use ' +
```
DominicKramer: nit: It could be helpful to declare a constant with the string's value and refer to the constant instead of the string literal.

src/agent/debuglet.ts
DominicKramer: I think we should say use `Stackdriver Debug Agent` (or something similar) instead of `debug agent` to make it clear that it is specifically the Stackdriver agent that is not impacted.

gaofanmichael: Since warning has been listened, once we reach this 'console.log' code, the warning was already fired and the message was printed. To be more concise, I changed the message to 'The Stackdriver debug agent does not use Inspector async stack traces. The INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE warning does not impact Stackdriver debug agent and can be ignored.'.

gaofanmichael: Done. see above changed message.

test/test-debuglet.ts

test/test-debuglet.ts
gaofanmichael: it has been used here and elsewhere. I have added a comment to fix this invalid method signature.

test/test-debuglet.ts

test/test-debuglet.ts

test/test-debuglet.ts
```diff
@@ -289,16 +291,26 @@ describe('Debuglet', function() {
...
+    });
 
     after(function() { process.env.GCLOUD_PROJECT = oldGP; });
-
-    beforeEach(function() {
+    let logText = '';
+    let oldLog: any;
+    beforeEach(() => {
       delete process.env.GCLOUD_PROJECT;
+      oldLog = console.log;
```
DominicKramer: Replacing `console.log` (and the associated restoration code) should no longer be needed, right?

test/test-debuglet.ts
```diff
@@ -315,6 +320,40 @@ describe('Debuglet', function() {
...
+      const scope = nock(API)
+                      .post(REGISTER_PATH)
+                      .reply(200, {debuggee: {id: DEBUGGEE_ID}});
+
+      debuglet.once('registered', function(id: string) {
+        assert.equal(id, DEBUGGEE_ID);
+        // TODO: Handle the case where debuglet.debuggee is undefined
+        assert.equal((debuglet.debuggee_ as Debuggee).project, projectId);
+        const arch = process.arch;
+        if (semver.satisfies(process.version, '>=8') &&
```
DominicKramer: nit: This should use `>=8.5` (or something similar) for the precise version of 8 that introduced the change.

src/agent/util/utils.ts
```diff
@@ -24,7 +24,10 @@ export const messages = {
   CAPTURE_BREAKPOINT_DATA: 'Error trying to capture snapshot data: ',
   INVALID_LINE_NUMBER: 'Invalid snapshot position: ',
   COULD_NOT_FIND_OUTPUT_FILE:
-      'Could not determine the output file associated with the transpiled input file'
+      'Could not determine the output file associated with the transpiled input file',
+  ASYNC_TRACES_WARNING: 'The Stackdriver debug agent does not use Inspector ' +
```
DominicKramer: nit: Maybe we should use:

`The Stackdriver Debugger for Node.js does not require V8 Inspector async stack traces. The INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE can be ignored.
`

test/test-debuglet.ts
```diff
@@ -289,16 +291,19 @@ describe('Debuglet', function() {
...
   describe('setup', function() {
-    before(function() { oldGP = process.env.GCLOUD_PROJECT; });
+    before(function() {
+      oldGP = process.env.GCLOUD_PROJECT;
+    });
 
     after(function() { process.env.GCLOUD_PROJECT = oldGP; });
-
-    beforeEach(function() {
+    beforeEach(() => {
```
DominicKramer: nit: For consistency with the rest of the file, we could use `function() {}` instead of `() => {}`.

test/test-debuglet.ts
```diff
@@ -289,16 +291,19 @@ describe('Debuglet', function() {
...
       delete process.env.GCLOUD_PROJECT;
       nocks.oauth2();
     });
 
-    afterEach(function() { nock.cleanAll(); });
+    afterEach(() => {
```
DominicKramer: nit: For consistency with the rest of the file, we could use `function() {}` instead of `() => {}`.

src/agent/util/utils.ts
gaofanmichael: Done.

test/test-debuglet.ts

test/test-debuglet.ts

PR #341, opened 2017-10-16
Support GCF through a 'initialization completed promise'

gaofanmichael:
This PR is to support Google Cloud Function (GCF) through initialization stackdriver debugger through a initialization of a promise and resolve the promise after breakpointfetch being scheduled.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -115,15 +115,15 @@ const formatBreakpoints = function(
 /**
  * CachedPromise class is designed to support debug agent in Google Cloud
  * Function (GCF). As GCF is a serverless environment and we wanted to make
- * sure debug agent could always capture the snapshots. We designed this class
- * so that GCF could wait the ListActiveBreakpoitns event completed before
- * throttling CPU to 0. CachedPromise has two member functions.
+ * sure debug agent could always capture the snapshots, we designed this class
+ * so that GCF could wait until the listActiveBreakpoints event is completed
+ * before throttling CPU to 0. CachedPromise has two member functions.
  * get() is to create or return an existing promise. resolve() will resolve
  * the promise at appropriate code location (in our case is after
- * listActiveBreakpoints executed).
+ * listActiveBreakpoints is executed).
  */
 export class CachedPromise {
-  private promise: Promise<void>;
+  private promise: Promise<void>|undefined;
   private promiseResolve: (() => void)|null;
   private promiseResolvedTimestamp = -Infinity;
   private refresh: number;
@@ -134,13 +134,11 @@ export class CachedPromise {
     const diff = Date.now() - this.promiseResolvedTimestamp;
     if (diff > this.refresh) {
       this.promise = new Promise<void>((resolve) => {
-        this.promiseResolve = () => {
-          resolve();
-        };
+        this.promiseResolve = resolve;
       });
     }
     assert(this.promise);
-    return this.promise;
+    return this.promise as Promise<void>;
   }
 
   resolve(): void {
@@ -160,7 +158,8 @@ export class Debuglet extends EventEmitter {
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
 
-  private cachedPromise: CachedPromise;
+  private cachedPromise =
+      new CachedPromise(PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS);
   // Exposed for testing
   config_: DebugAgentConfig;
   fetcherActive_: boolean;
@@ -221,9 +220,6 @@ export class Debuglet extends EventEmitter {
 
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap_ = {};
-
-    this.cachedPromise =
-        new CachedPromise(PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS);
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
@@ -372,7 +368,7 @@ export class Debuglet extends EventEmitter {
 
     });
   }
-  isReady() {
+  isReady(): Promise<void> {
     return this.cachedPromise.get();
   }
 
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -113,32 +113,29 @@ const formatBreakpoints = function(
 };
 
 /**
- * CachedPromise class is designed to support debug agent in Google Cloud
- * Function (GCF). As GCF is a serverless environment and we wanted to make
- * sure debug agent could always capture the snapshots, we designed this class
- * so that GCF could wait until the listActiveBreakpoints event is completed
- * before throttling CPU to 0. CachedPromise has two member functions.
- * get() is to create or return an existing promise. resolve() will resolve
- * the promise at appropriate code location (in our case is after
- * listActiveBreakpoints is executed).
+ * CachedPromise stores a promise for a limited time. Its member function get()
+ * will initially create a promise, or create a promise when previous promise
+ * stales, then return that promise. resolve() will resolve the stored promise.
  */
 export class CachedPromise {
-  private promise: Promise<void>|undefined;
+  private promise: Promise<void>;
   private promiseResolve: (() => void)|null;
   private promiseResolvedTimestamp = -Infinity;
-  private refresh: number;
-  constructor(refresh: number) {
-    this.refresh = refresh;
+  private timeUntilStaleMS: number;
+  constructor(timeUntilStaleMS: number) {
+    this.timeUntilStaleMS = timeUntilStaleMS;
+    this.promise = new Promise<void>((resolve) => {
+      this.promiseResolve = resolve;
+    });
   }
   get(): Promise<void> {
     const diff = Date.now() - this.promiseResolvedTimestamp;
-    if (diff > this.refresh) {
+    if (diff > this.timeUntilStaleMS) {
       this.promise = new Promise<void>((resolve) => {
         this.promiseResolve = resolve;
       });
     }
-    assert(this.promise);
-    return this.promise as Promise<void>;
+    return this.promise;
   }
 
   resolve(): void {
@@ -368,6 +365,15 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+
+  /**
+   * isReady is designed to support debug agent on Google Cloud Function (GCF).
+   * As GCF is a serverless environment and we wanted to make sure debug agent
+   * always captures the snapshots, we used CachedPromise so that GCF could wait
+   * until the listActiveBreakpoints event is completed before throttling CPU to
+   * 0. Here the cachedPromise will be resolved after listActiveBreakpoints is
+   * executed).
+   */
   isReady(): Promise<void> {
     return this.cachedPromise.get();
   }
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -125,7 +125,7 @@ const formatBreakpoints = function(
 export class CachedPromise {
   private promise: Promise<void>;
   private promiseResolve: (() => void)|null;
-  private promiseResolvedTimestamp: number = -Infinity;
+  private promiseResolvedTimestamp = -Infinity;
   private refresh: number;
   constructor(refresh: number) {
     this.refresh = refresh;
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -62,10 +62,11 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
-// PROMISE_RESOLVE_CUT_OFF_IN_SECONDS is a heuristic that we set to force the
-// debug agent return a promise when checkReady is called. The value is selected
-// between Stackdriver debugger hanging get duration (40s) and TCP default
-// time-out (https://tools.ietf.org/html/rfc5482).
+// PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS is a heuristic that we set to force
+// the debug agent to return a promise when isReady is called in
+// isReadyManager. The value is selected between Stackdriver debugger hanging
+// get duration (40s) and TCP default time-out
+// (https://tools.ietf.org/html/rfc5482).
 const PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS = 60 * 1000;
 
 /**
@@ -112,6 +113,45 @@ const formatBreakpoints = function(
           .join('\n');
 };
 
+/**
+ * isReadyManager class is designed to support debug agent in Google Cloud
+ * Function (GCF). As GCF is a serverless environment and we wanted to make
+ * sure debug agent could always capture the snapshots. We designed this class
+ * so that GCF could wait the ListActiveBreakpoitns event completed before
+ * throttling CPU to 0. isReadyManager has two member functions.
+ * isReady is to create a initial promise, or create a promise to replace the
+ * old one if it steals. makeReady will resolve the promise at appropriate
+ * code location (in our case is after listActiveBreakpoints executed).
+ */
+export class IsReadyManager {
+  private promise: Promise<void>;
+  private promiseResolve: (() => void)|null;
+  private promiseResolvedTimestamp: number;
+  constructor() {
+    this.promiseResolvedTimestamp = -Infinity;
+  }
+  isReady() {
+    const diff = Date.now() - this.promiseResolvedTimestamp;
+    if (diff > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      this.promise = new Promise<void>((resolve) => {
+        this.promiseResolve = () => {
+          resolve();
+        };
+      });
+    }
+    assert(this.promise);
+    return this.promise;
+  }
+
+  makeReady() {
+    this.promiseResolvedTimestamp = Date.now();
+    if (this.promiseResolve) {
+      this.promiseResolve();
+      this.promiseResolve = null;
+    }
+  }
+}
+
 export class Debuglet extends EventEmitter {
   private debug_: Debug;
   private v8debug_: DebugApi|null;
@@ -120,9 +160,7 @@ export class Debuglet extends EventEmitter {
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
 
-  private promise: Promise<void>;
-  private promiseResolve: (() => void)|null;
-  private promiseResolvedTimestamp: number;
+  private isReadyManager: IsReadyManager;
   // Exposed for testing
   config_: DebugAgentConfig;
   fetcherActive_: boolean;
@@ -184,7 +222,7 @@ export class Debuglet extends EventEmitter {
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap_ = {};
 
-    this.promiseResolvedTimestamp = -Infinity;
+    this.isReadyManager = new IsReadyManager();
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
@@ -333,16 +371,8 @@ export class Debuglet extends EventEmitter {
 
     });
   }
-  checkReady() {
-    const diff = Date.now() - this.promiseResolvedTimestamp;
-    if (diff > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
-      this.promise = new Promise<void>((resolve) => {
-        this.promiseResolve = () => {
-          resolve();
-        };
-      });
-    }
-    return this.promise;
+  isReady() {
+    return this.isReadyManager.isReady();
   }
 
   /**
@@ -619,11 +649,7 @@ export class Debuglet extends EventEmitter {
                 }
                 that.scheduleBreakpointFetch_(
                     that.config_.breakpointUpdateIntervalSec);
-                that.promiseResolvedTimestamp = Date.now();
-                if (that.promiseResolve) {
-                  that.promiseResolve();
-                  that.promiseResolve = null;
-                }
+                that.isReadyManager.makeReady();
                 return;
             }
           });
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -116,8 +116,8 @@ export class Debuglet extends EventEmitter {
   private completedBreakpointMap_: {[key: string]: boolean};
 
   private promise_: Promise<void>|null;
-  private promiseResolve_: (value?: void|PromiseLike<void>|undefined) => void;
-  private promiseResolvedTimestamp_: [number, number];
+  private promiseResolve_: () => void;
+  private promiseResolvedTimestamp_: number;
   // Exposed for testing
   config_: DebugAgentConfig;
   fetcherActive_: boolean;
@@ -180,7 +180,7 @@ export class Debuglet extends EventEmitter {
     this.completedBreakpointMap_ = {};
 
     /** @private {Array<number>} */
-    this.promiseResolvedTimestamp_ = [0, 0];
+    this.promiseResolvedTimestamp_ = Date.now();
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
@@ -330,8 +330,8 @@ export class Debuglet extends EventEmitter {
     });
   }
   checkReady() {
-    const diff = process.hrtime(this.promiseResolvedTimestamp_);
-    if (diff[0] < PROMISE_RESOLVE_CUT_OFF_IN_SECONDS) {
+    const diff = Date.now() - this.promiseResolvedTimestamp_;
+    if (diff < PROMISE_RESOLVE_CUT_OFF_IN_SECONDS * 1000) {
       return Promise.resolve();
     }
     this.promise_ = new Promise<void>((resolve) => {
@@ -618,7 +618,7 @@ export class Debuglet extends EventEmitter {
                     that.config_.breakpointUpdateIntervalSec);
                 if (that.promise_) {
                   that.promiseResolve_();
-                  that.promiseResolvedTimestamp_ = process.hrtime();
+                  that.promiseResolvedTimestamp_ = Date.now();
                   that.promise_ = null;
                 }
                 return;
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -327,7 +327,7 @@ export class Debuglet extends EventEmitter {
 
     });
   }
-  initializationPromise() {
+  initializePromise() {
     this.promiseInitialized_ = true;
     return new Promise<void>((resolve) => {
       this.promiseResolve_ = () => {
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -562,7 +562,6 @@ export class Debuglet extends EventEmitter {
                   that.config_.internal.registerDelayOnFetcherErrorSec);
               return;
             }
-
             // TODO: Address the case where `response` is `undefined`.
             switch ((response as http.ServerResponse).statusCode) {
               case 404:
@@ -610,14 +609,14 @@ export class Debuglet extends EventEmitter {
                 }
                 that.scheduleBreakpointFetch_(
                     that.config_.breakpointUpdateIntervalSec);
+                if (that.promiseInitialized_) {
+                  that.promiseResolve_();
+                  that.promiseInitialized_ = false;
+                }
                 return;
             }
           });
     }, seconds * 1000).unref();
-    if (this.promiseInitialized_) {
-      this.promiseResolve_();
-      this.promiseInitialized_ = false;
-    }
   }
 
   /**
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -61,6 +61,7 @@ const NODE_VERSION_MESSAGE =
 const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
+const PROMISE_RESOLVE_CUT_OFF_IN_SECONDS = 30;
 
 /**
  * Formats a breakpoint object prefixed with a provided message as a string
@@ -113,9 +114,10 @@ export class Debuglet extends EventEmitter {
   private project_: string|null;
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
-  private promiseInitialized_: boolean;
-  private promiseResolve_: (value?: void|PromiseLike<void>|undefined) => void;
 
+  private promise_: Promise<void>|null;
+  private promiseResolve_: (value?: void|PromiseLike<void>|undefined) => void;
+  private promiseResolvedTimestamp_: [number, number];
   // Exposed for testing
   config_: DebugAgentConfig;
   fetcherActive_: boolean;
@@ -177,8 +179,8 @@ export class Debuglet extends EventEmitter {
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap_ = {};
 
-    /** @private {boolean} */
-    this.promiseInitialized_ = false;
+    /** @private {Array<number>} */
+    this.promiseResolvedTimestamp_ = [0, 0];
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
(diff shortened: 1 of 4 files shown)


src/agent/debuglet.ts
```diff
@@ -314,6 +319,14 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+  initializationPromise() {
```
kjin: How will this function be used? I don't see a call to it anywhere.

src/agent/debuglet.ts
```diff
@@ -314,6 +319,14 @@ export class Debuglet extends EventEmitter {
...
+    this.promiseInitialized_ = true;
+    return new Promise<void>((resolve) => {
+      this.promiseResolve_ = () => {
```
kjin: If `initializationPromise` is called twice, `this.promiseResolve_` will have its previous value overwritten, and the Promise returned in the first call will never be resolved. I suppose it depends on the usage of `initializationPromise`, but this seems like it might not be what we want.

src/agent/debuglet.ts
```diff
@@ -115,6 +115,8 @@ export class Debuglet extends EventEmitter {
   private project_: string|null;
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
+  private promiseInitialized_: boolean;
+  private promiseResolve_: (value?: void|PromiseLike<void>|undefined) => void;
```
kjin: @gaofanmichael The way Promises are being used, it looks like an event emission might be more suitable (anyone who needs to wait for a Promise returned by `initializationPromise` to resolve can just as easily listen for an `"initialized"` event.) Is there a reason in particular we need to use a Promise?

gaofanmichael: Do you mean as below? 
debug.initializePromise();
process.on('breakpointfetched', () => {})

src/agent/debuglet.ts
```diff
@@ -117,6 +118,9 @@ export class Debuglet extends EventEmitter {
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
 
+  private promise_: Promise<void>|null;
+  private promiseResolve_: (value?: void|PromiseLike<void>|undefined) => void;
```
kjin: Is this the right type? When you call `this.promiseResolve_`, you don't pass any arguments in.

src/agent/debuglet.ts
```diff
@@ -322,6 +329,18 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+  checkReady() {
+    const diff = process.hrtime(this.promiseResolvedTimestamp_);
```
kjin: Is `process.hrtime` necessary? We only need timestamp in seconds. (Not sure what the best practice is when dealing with timescales on the order of seconds.)

src/agent/debuglet.ts
```diff
@@ -322,6 +329,18 @@ export class Debuglet extends EventEmitter {
...
+    if (diff[0] < PROMISE_RESOLVE_CUT_OFF_IN_SECONDS) {
+      return Promise.resolve();
+    }
+    this.promise_ = new Promise<void>((resolve) => {
```
kjin: I understand that GCF is restricted to calling `checkReady` a second time only after the Promise returned by the first call has been resolved... buthis is not a general restriction -- users can still call `checkReady` arbitrarily. Given this, `this.promiseResolve_` may have its value clobbered. (Also see: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/341#discussion_r145486623)

test/test-debuglet.ts
```diff
@@ -871,6 +871,34 @@ describe('Debuglet', function() {
...
+                      .reply(200, {debuggee: {id: DEBUGGEE_ID}})
+                      .get(BPS_PATH + '?successOnTimeout=true')
+                      .reply(200, {breakpoints: []});
+      debuglet.start();
+      const debugPromise = debuglet.checkReady();
+
+      debuglet.once('registered', function reg(id: string) {
+        assert.equal(id, DEBUGGEE_ID);
+        debugPromise.then(() => {
+          setTimeout(function() {
```
kjin: Is `setTimeout` necessary here?

src/agent/debuglet.ts
```diff
@@ -175,6 +179,9 @@ export class Debuglet extends EventEmitter {
 
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap_ = {};
+
+    /** @private {Array<number>} */
```
jinwoo: Why are we putting this JsDoc everywhere when this is typescript? Maybe because this code has been migrated from JS?

And also why all those trailing underscores in private member names?

src/agent/debuglet.ts
```diff
@@ -175,6 +179,9 @@ export class Debuglet extends EventEmitter {
...
+    this.promiseResolvedTimestamp_ = [0, 0];
```
jinwoo: Per the Node document, passing a user-defined array to process.hrtime() is an undefined behavior. I think you should initialize this with the result of process.hrtime().

src/agent/debuglet.ts
jinwoo: +1. process.hrtime() is only for high-resolution time data. We don't need that fine granular data, do we?

gaofanmichael: I used high resolution time here as in legacy-debugapi.ts we have measured capture time using hrtime. I agree that here, measuring time difference at around 30 seconds, we can use Date.now() instead.

test/test-debuglet.ts
gaofanmichael: I think so. This is a format to make sure nock.done is called to check that all specified calls on that scope were performed.

src/agent/debuglet.ts
gaofanmichael: This type signature was the signature for promise.resolve.
Reconsidering this I think you are right, as we do not necessary to follow the signature for promise.resolve().

I updated this signature to be () => void.

src/agent/debuglet.ts
gaofanmichael: First question -- yes 
Regarding the names, actually another PR is addressing right now. 
https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/344

I will remove the trailing underscore for this variable here.

src/agent/debuglet.ts
gaofanmichael: Good question :)
Per our discussion earlier today, we will return the same promise if the cached promise existed. Otherwise, we will calculate the time since last listBreakpoint and determine if resolve a promise immediately or create a new promise and cache it.

src/agent/debuglet.ts
gaofanmichael: I change hrtime to Date.now() and will initialize with Date.now()

src/agent/debuglet.ts
```diff
@@ -62,6 +62,12 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
+// PROMISE_RESOLVE_CUT_OFF_IN_SECONDS is a heuristic that we set to force the
+// debug agent return a promise when checkReady is called. The value is selected
+// between Stackdriver debugger hanging get duration (40s) and TCP default
+// time-out (https://tools.ietf.org/html/rfc5482).
+const PROMISE_RESOLVE_CUT_OFF_IN_SECONDS = 60;
```
kjin: Nit: Make this _IN_MILLISECONDS and have its value set to 60 * 1000

src/agent/debuglet.ts
```diff
@@ -174,6 +183,9 @@ export class Debuglet extends EventEmitter {
 
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap_ = {};
+
+    /** @private {Array<number>} */
+    this.promiseResolvedTimestamp_ = Date.now();
```
kjin: Can you set this value to either be `PROMISE_RESOLVE_CUT_OFF_IN_SECONDS` less than `Date.now()`, or some sentinel value? Otherwise Promises returned by `checkReady` when it's called within `PROMISE_RESOLVE_CUT_OFF_IN_SECONDS` within initialization will be incorrectly resolved.

src/agent/debuglet.ts
```diff
@@ -322,6 +334,19 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+  checkReady() {
+    if (this.promise_) return this.promise_;
+    const diff = Date.now() - this.promiseResolvedTimestamp_;
+    if (diff < PROMISE_RESOLVE_CUT_OFF_IN_SECONDS * 1000) {
+      return Promise.resolve();
```
jinwoo: I think you can always return `this.promise_` no matter it's resolved or not, instead of resetting `this.promise_` to `null` at line 627 and returning `Promise.resolve()` from here.

There are three cases:

1. The first time `checkReady()` is called (`this.promise_` is `null`):
   You'll have to unconditionally create a new promise.
2. When `diff` < cutoff threshold:
   Just return `this.promise_`.
3. When `diff` > cutoff thresdhold:
   Create a new promise.

src/agent/debuglet.ts
```diff
@@ -597,6 +622,11 @@ export class Debuglet extends EventEmitter {
                 }
                 that.scheduleBreakpointFetch_(
                     that.config_.breakpointUpdateIntervalSec);
+                if (that.promise_) {
```
jinwoo: For my comment above to work, this check should be `if (that.promiseResolve_)` ...

src/agent/debuglet.ts
```diff
@@ -597,6 +622,11 @@ export class Debuglet extends EventEmitter {
...
+                  that.promiseResolve_();
+                  that.promise_ = null;
```
jinwoo: ... and this should be `that.promiseResolve_ = null;`

src/agent/debuglet.ts
jinwoo: Please delete this JsDoc (or at least make it correct). It's not a number array. Bonus point if you delete the underscore too :)

src/agent/debuglet.ts
jinwoo: I think it's best to set this to `-Infinity` (or 0).

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
```diff
@@ -62,6 +62,12 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
+// PROMISE_RESOLVE_CUT_OFF_IN_SECONDS is a heuristic that we set to force the
```
jinwoo: Update the variable name here too.

src/agent/debuglet.ts
```diff
@@ -62,6 +62,12 @@ const BREAKPOINT_ACTION_MESSAGE =
...
+// debug agent return a promise when checkReady is called. The value is selected
```
jinwoo: Nit: s/return/to return

src/agent/debuglet.ts
```diff
@@ -322,6 +333,17 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+  checkReady() {
+    const diff = Date.now() - this.promiseResolvedTimestamp;
+    if (diff > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
```
jinwoo: I think this should be

```javascript
  if (!this.promise || diff > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
```

src/agent/debuglet.ts
```diff
@@ -114,6 +120,9 @@ export class Debuglet extends EventEmitter {
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
 
+  private promise: Promise<void>;
```
DominicKramer: As another point of view, you could also make a `IsReadyManager` (or something similar) class that contains these few properties and exposes `isReady` and `resolvePromise` methods (or use some better names :)).  That way those methods could be documented and tested in isolation and the `Debuglet` would just use an instance of `IsReadyManager`.  This would allow for separation of concerns.  Its just a different way of looking at the problem.

kjin: +1; at the very least `promise*` should be changed to something more descriptive.

src/agent/debuglet.ts
gaofanmichael: Understood. but since we set the initial this.promiseResolvedTimestamp to be -Infinite. We can be sure that the first diff will always > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS, right?

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
jinwoo: Makes sense. But since that contract can easily be broken in the future, what about adding an assert at line 345 that `this.promise` has some value?

src/agent/debuglet.ts
```diff
@@ -112,6 +113,45 @@ const formatBreakpoints = function(
...
+ * isReady is to create a initial promise, or create a promise to replace the
+ * old one if it steals. makeReady will resolve the promise at appropriate
+ * code location (in our case is after listActiveBreakpoints executed).
+ */
+export class IsReadyManager {
+  private promise: Promise<void>;
+  private promiseResolve: (() => void)|null;
+  private promiseResolvedTimestamp: number;
+  constructor() {
+    this.promiseResolvedTimestamp = -Infinity;
```
jinwoo: Nit: you can initialize just in line 129 above and delete this constructor.

src/agent/debuglet.ts
```diff
@@ -112,6 +113,45 @@ const formatBreakpoints = function(
...
+  }
+  isReady() {
```
jinwoo: It'd be good to explicitly specify the function return type for all the public functions. That's good for documentation (when the user of this class reads the code) and also for preventing from breaking existing interface accidentally.

src/agent/debuglet.ts
```diff
@@ -112,6 +113,45 @@ const formatBreakpoints = function(
...
+        this.promiseResolve = () => {
+          resolve();
+        };
+      });
+    }
+    assert(this.promise);
+    return this.promise;
+  }
+
+  makeReady() {
```
jinwoo: What about `markReady()`?

src/agent/debuglet.ts
```diff
@@ -112,6 +113,45 @@ const formatBreakpoints = function(
...
+ * isReadyManager class is designed to support debug agent in Google Cloud
+ * Function (GCF). As GCF is a serverless environment and we wanted to make
+ * sure debug agent could always capture the snapshots. We designed this class
+ * so that GCF could wait the ListActiveBreakpoitns event completed before
+ * throttling CPU to 0. isReadyManager has two member functions.
+ * isReady is to create a initial promise, or create a promise to replace the
+ * old one if it steals. makeReady will resolve the promise at appropriate
+ * code location (in our case is after listActiveBreakpoints executed).
+ */
+export class IsReadyManager {
```
jinwoo: I don't like the name `IsReadyManager` as a class name. It sounds like a function name. But I don't have a good idea either.

src/agent/debuglet.ts
```diff
@@ -62,6 +62,13 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
+// PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS is a heuristic that we set to force
+// the debug agent to return a promise when isReady is called in
+// isReadyManager. The value is selected between Stackdriver debugger hanging
+// get duration (40s) and TCP default time-out
+// (https://tools.ietf.org/html/rfc5482).
+const PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS = 60 * 1000;
```
ofrobots: Why is this 60 seconds and not 100 seconds. The comment should explain why this number is reasonable. It mentions that it should be somewhere in the range.. but where in the range

src/agent/debuglet.ts
```diff
@@ -62,6 +62,13 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
+// PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS is a heuristic that we set to force
+// the debug agent to return a promise when isReady is called in
+// isReadyManager. The value is selected between Stackdriver debugger hanging
+// get duration (40s) and TCP default time-out
```
ofrobots: Food for thought: perhaps on GCF we should not using a hanging GET.

src/agent/debuglet.ts
```diff
@@ -106,6 +113,45 @@ const formatBreakpoints = function(
...
+  private promise: Promise<void>;
+  private promiseResolve: (() => void)|null;
+  private promiseResolvedTimestamp: number;
+  constructor() {
+    this.promiseResolvedTimestamp = -Infinity;
+  }
+  isReady() {
+    const diff = Date.now() - this.promiseResolvedTimestamp;
+    if (diff > PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      this.promise = new Promise<void>((resolve) => {
```
ofrobots: Could be simpler:

```ts
this.promise = new Promise((resolve) { this.resolve = resolve; }
```

That is, you don't need the extra function that simply calls `resolve`.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
```diff
@@ -106,6 +112,46 @@ const formatBreakpoints = function(
...
+ * Function (GCF). As GCF is a serverless environment and we wanted to make
+ * sure debug agent could always capture the snapshots. We designed this class
+ * so that GCF could wait the ListActiveBreakpoitns event completed before
+ * throttling CPU to 0. CachedPromise has two member functions.
+ * get() is to create or return an existing promise. resolve() will resolve
+ * the promise at appropriate code location (in our case is after
+ * listActiveBreakpoints executed).
+ */
+export class CachedPromise {
+  private promise: Promise<void>;
```
jinwoo: I'm not comfortable with this being declared as `Promise<void>` and not initialized. What about making it `Promise<void>|undefined`?

src/agent/debuglet.ts
```diff
@@ -106,6 +112,46 @@ const formatBreakpoints = function(
           .join('\n');
 };
 
+/**
+ * CachedPromise class is designed to support debug agent in Google Cloud
+ * Function (GCF). As GCF is a serverless environment and we wanted to make
+ * sure debug agent could always capture the snapshots. We designed this class
```
jinwoo: Nit. "...snapshot. We designed..." -> "...snapshot, we designed..."

src/agent/debuglet.ts
```diff
@@ -106,6 +112,46 @@ const formatBreakpoints = function(
...
+ * so that GCF could wait the ListActiveBreakpoitns event completed before
```
jinwoo: Nit.

ListActiveBreakpoitns -> ListActiveBreakpoints
"wait" -> "wait until"
"event completed" -> "event is completed"

src/agent/debuglet.ts
```diff
@@ -106,6 +112,46 @@ const formatBreakpoints = function(
...
+ * throttling CPU to 0. CachedPromise has two member functions.
+ * get() is to create or return an existing promise. resolve() will resolve
+ * the promise at appropriate code location (in our case is after
+ * listActiveBreakpoints executed).
```
jinwoo: More nit: listActiveBreakpoints *is* executed

BTW is it *l*istActiveBreakpoints or *L*istActiveBreakpoints?

src/agent/debuglet.ts
```diff
@@ -106,6 +112,46 @@ const formatBreakpoints = function(
...
+  constructor(refresh: number) {
+    this.refresh = refresh;
+  }
+  get(): Promise<void> {
+    const diff = Date.now() - this.promiseResolvedTimestamp;
+    if (diff > this.refresh) {
+      this.promise = new Promise<void>((resolve) => {
+        this.promiseResolve = () => {
+          resolve();
+        };
```
jinwoo: Making this a one-liner would be simpler.

```typescript
this.promiseResolve = resolve;
```

src/agent/debuglet.ts
```diff
@@ -114,6 +160,7 @@ export class Debuglet extends EventEmitter {
   private controller_: Controller;
   private completedBreakpointMap_: {[key: string]: boolean};
 
+  private cachedPromise: CachedPromise;
```
jinwoo: Just initialize here instead of from the constructor.

src/agent/debuglet.ts
```diff
@@ -322,6 +372,9 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+  isReady() {
```
jinwoo: Please explicitly declare the return type.

test/test-debuglet.ts
```diff
@@ -74,6 +74,19 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
   const hasCorrectDescription = status.description.format.match(re);
   return status.isError && hasCorrectDescription;
 }
+describe('CachedPromise', function() {
```
jinwoo: Arrow function?

test/test-debuglet.ts
```diff
@@ -74,6 +74,19 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
...
+  it('CachedPromise can proceed execute when resolve is called', function(done) {
```
jinwoo: Arrow function here too?

test/test-debuglet.ts
```diff
@@ -74,6 +74,19 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
...
+    this.timeout(2000);
+    let cachedPromise = new CachedPromise(60 * 1000);
+    let promise = cachedPromise.get();
+    setTimeout(function(){
```
jinwoo: And here too.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done. it is listActiveBreakpoints.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

test/test-debuglet.ts

test/test-debuglet.ts

test/test-debuglet.ts

src/agent/debuglet.ts
```diff
@@ -106,6 +112,44 @@ const formatBreakpoints = function(
...
+  }
+  get(): Promise<void> {
+    const diff = Date.now() - this.promiseResolvedTimestamp;
+    if (diff > this.refresh) {
+      this.promise = new Promise<void>((resolve) => {
+        this.promiseResolve = resolve;
+      });
+    }
+    assert(this.promise);
+    return this.promise as Promise<void>;
```
jinwoo: I wish typescript understands `assert()` as a type guard. If you change `assert(this.promise)` to something like `if (!this.promise) throw new Error('error!!');`, typescript would know that `this.promise` is defined at this point.

Just a rant. Maybe I should file a feature request on typescript.

jinwoo: Filed one: https://github.com/Microsoft/TypeScript/issues/19622

src/agent/debuglet.ts
```diff
@@ -106,6 +112,44 @@ const formatBreakpoints = function(
...
+ * before throttling CPU to 0. CachedPromise has two member functions.
+ * get() is to create or return an existing promise. resolve() will resolve
+ * the promise at appropriate code location (in our case is after
+ * listActiveBreakpoints is executed).
+ */
+export class CachedPromise {
+  private promise: Promise<void>|undefined;
+  private promiseResolve: (() => void)|null;
+  private promiseResolvedTimestamp = -Infinity;
+  private refresh: number;
```
kjin: The name of this could be clearer... maybe `timeUntilStaleMS`?

src/agent/debuglet.ts
```diff
@@ -106,6 +112,44 @@ const formatBreakpoints = function(
...
+  constructor(refresh: number) {
+    this.refresh = refresh;
```
kjin: If you initialize `this.promise` here, its type can just be `Promise<void>` and you don't need the `assert`/cast in `get()`.

src/agent/debuglet.ts
```diff
@@ -106,6 +112,44 @@ const formatBreakpoints = function(
           .join('\n');
 };
 
+/**
+ * CachedPromise class is designed to support debug agent in Google Cloud
```
kjin: IMO this description fits `Debuggee#isReady` better than `CachedPromise`. `CachedPromise` doesn't have anything to do with GCF on its own; it's a general-purpose construct that allows us to wait for a locally cached resource to be refreshed if it's deemed stale.

src/debuggee.ts
```diff
@@ -34,10 +34,10 @@ export interface DebuggeeProperties {
 }
 
 export class Debuggee {
-  private uniquifier: string;
-  private description: string;
-  private agentVersion?: string;
-  private sourceContexts?: Array<{[key: string]: any}>;
+  uniquifier: string;
```
kjin: 👍

test/test-debuglet.ts
```diff
@@ -75,6 +75,20 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
   return status.isError && hasCorrectDescription;
 }
 
+describe('CachedPromise', function() {
+  it('CachedPromise can proceed execute when resolve is called', (done) => {
+    this.timeout(2000);
+    let cachedPromise = new CachedPromise(60 * 1000);
```
kjin: It might be more comprehensive to let the CachedPromise go stale as part of the test.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

test/test-debuglet.ts

src/agent/debuglet.ts
gaofanmichael: Reorganized the comments.

src/agent/debuglet.ts
kjin: FYI @jinwoo -- didn't see your comment earlier, but I advocated for keeping it as `Promise<void>` and having it initialized instead.

src/agent/debuglet.ts
```diff
@@ -368,6 +365,15 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+
+  /**
+   * isReady is designed to support debug agent on Google Cloud Function (GCF).
+   * As GCF is a serverless environment and we wanted to make sure debug agent
+   * always captures the snapshots, we used CachedPromise so that GCF could wait
```
kjin: Nit: The user doesn't know about `CachedPromise` (nor should they) -- to them, they should call this function and expect to get a `Promise` that's resolved when they know that the list of breakpoints is no longer stale. Can you re-frame this explanation to account for this? Keep in mind that this is on the public API of the Debug Agent now.

src/agent/debuglet.ts
jinwoo: @kjin I think that's better.

test/test-debuglet.ts
```diff
@@ -87,6 +87,28 @@ describe('CachedPromise', function() {
       done();
     });
   });
+
+  it('CachedPromise should return the same promise if not staled', (done) => {
```
kjin: I don't think it's part of the public contract of `CachedPromise` to return same/difference `Promise` objects based on staleness. What I would expect of `CachedPromise` instead, is that if the underlying resource is deemed "fresh", then the `Promise` object we get from `CachedPromise#get` is already resolved; otherwise, it will not be resolved until `CachedPromise#resolve` is called. I think this is the behavior that should be tested... what do you think?

src/agent/debuglet.ts
```diff
@@ -113,32 +113,29 @@ const formatBreakpoints = function(
...
- * so that GCF could wait until the listActiveBreakpoints event is completed
- * before throttling CPU to 0. CachedPromise has two member functions.
- * get() is to create or return an existing promise. resolve() will resolve
- * the promise at appropriate code location (in our case is after
- * listActiveBreakpoints is executed).
+ * CachedPromise stores a promise for a limited time. Its member function get()
+ * will initially create a promise, or create a promise when previous promise
+ * stales, then return that promise. resolve() will resolve the stored promise.
  */
 export class CachedPromise {
```
kjin: Does `CachedPromise` need to be exported?

src/agent/debuglet.ts
```diff
@@ -368,6 +365,15 @@ export class Debuglet extends EventEmitter {
 
     });
   }
+
+  /**
+   * isReady is designed to support debug agent on Google Cloud Function (GCF).
```
kjin: Nit: The user doesn't know about CachedPromise (nor should they) -- to them, they should call this function and expect to get a Promise that's resolved when they know that the list of breakpoints is no longer stale. Can you re-frame this explanation to account for this? Keep in mind that this is on the public API of the Debug Agent now.

src/agent/debuglet.ts
gaofanmichael: It is exported for testing.

src/debuggee.ts
DominicKramer: Why are these no longer private?

src/agent/debuglet.ts
gaofanmichael: Done.

test/test-debuglet.ts

src/debuggee.ts
gaofanmichael: Current compiler will complain their values were never read. However, we cannot remove them as they need to be send to debug server in wire. Since it needs to be exposed to debug server anyway, I made them public.

kjin: `Debuggee` might be better off as an interface, but that's out of the scope of this PR.

test/test-debuglet.ts
```diff
@@ -75,6 +75,53 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
   return status.isError && hasCorrectDescription;
 }
 
+describe('CachedPromise', function() {
+  it('CachedPromise can proceed execute when resolve is called', (done) => {
```
kjin: This test is already covered by `Stale promise...` and can be removed.

src/agent/debuglet.ts
```diff
@@ -324,6 +367,18 @@ export class Debuglet extends EventEmitter {
   }
 
   /**
+   * isReady is designed to support debug agent on Google Cloud Function (GCF).
```
kjin: Nit: I believe @DominicKramer mentioned this earlier, but `isReady` is not specific to GCF (even though there are no known additional use cases). Instead of describing what it is used for, I would recommend limiting this comment to describing its expected behavior.

gaofanmichael:
Opened another PR https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/358
Close this one.

PR #343, opened 2017-10-18
Inspector only enabled when specified

gaofanmichael:
This PR ensures that inspector protocol will only be enabled if
environmental variable GCLOUD_USE_INSPECTOR is set.


src/agent/v8/debugapi.ts
```diff
@@ -46,10 +46,15 @@ const nodeVersion = /v(\d+\.\d+\.\d+)/.exec(process.version);
...
   const dummyapi = require('./dummy-debugapi');
   debugApiConstructor = dummyapi.DummyDebugApi;
-} else if (semver.satisfies(nodeVersion[1], '>=8')) {
+} else if (
+    semver.satisfies(nodeVersion[1], '>=8') &&
+    process.env.GCLOUD_USE_INSPECTOR) {
   const inspectorapi = require('./inspector-debugapi');
   debugApiConstructor = inspectorapi.InspectorDebugApi;
 } else {
+  if (process.env.GCLOUD_USE_INSPECTOR) {
```
kjin: Maybe we should move this check to the V8DebugAPI constructor, because there we have access to a logger object. @gaofanmichael @DominicKramer thoughts?

bin/run-test.sh
```diff
@@ -32,6 +32,14 @@ function run {
 # Run test/coverage
 run build/test
 
+node_version_major=$(node -v | sed 's/v\([0-9]*\)\.[0-9]*.[0-9]*/\1/')
```
kjin: I recommend
```sh
node_version_major=$(node -p "require('semver').major(process.version)")
```

test_windows.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+});
+
+process.env.GCLOUD_PROJECT = 0
+process.env.CLOUD_DEBUG_ASSERTIONS = 1
+
+const nodeVersion = /v(\d+\.\d+\.\d+)/.exec(process.version);
+if (!nodeVersion || nodeVersion.length < 2) {
+  console.error('unable to get node version');
+} else if (semver.satisfies(nodeVersion[1], '>=8')) {
```
kjin: I understand this is the same logic as in `debuglet.ts` but I think it's unnecessary to parse the version string beforehand. `semver.satisfies(process.version, '>=8')` should just work here, and we don't need the first condition (that might even mask the issue because it doesn't exit with code 1).

test_windows.js
```diff
@@ -0,0 +1,31 @@
+var Mocha = require('mocha')
+var fs = require('fs');
+var path = require('path');
+var semver = require('semver');
+
+
+var mocha = new Mocha();
+var testDir = 'build/test';
+fs.readdirSync(testDir).filter(function(file) {
+    return file.substr(-3) === '.js';
```
kjin: `file.endsWith('.js')`

src/agent/v8/debugapi.ts
DominicKramer: I agree

test_windows.js
```diff
@@ -0,0 +1,31 @@
+var Mocha = require('mocha')
+var fs = require('fs');
+var path = require('path');
+var semver = require('semver');
+
+
+var mocha = new Mocha();
+var testDir = 'build/test';
```
DominicKramer: I suggest using `path.join('build', 'test')` since Windows uses a different path separator.

test_windows.js
```diff
@@ -0,0 +1,31 @@
...
+    return file.substr(-3) === '.js';
+}).forEach(function(file) {
+    mocha.addFile(
+        path.join(testDir, file)
+    );
+});
+
+process.env.GCLOUD_PROJECT = 0
+process.env.CLOUD_DEBUG_ASSERTIONS = 1
+
```
DominicKramer: Don't you want to call `mocha` here without `GCLOUD_USE_INSPECTOR` set to run the tests without using inspector?

gaofanmichael: I find it hard to run both in one js file. So I leave that in appveyor and have this js just for inspector.

bin/run-test.sh
gaofanmichael: Done.

test_windows.js
gaofanmichael: Done.

test_windows.js
gaofanmichael: Sounds good. I will this elsewhere.

test_windows.js
gaofanmichael: Done.

src/agent/util/utils.ts
```diff
@@ -24,7 +24,8 @@ export const messages = {
   CAPTURE_BREAKPOINT_DATA: 'Error trying to capture snapshot data: ',
   INVALID_LINE_NUMBER: 'Invalid snapshot position: ',
   COULD_NOT_FIND_OUTPUT_FILE:
-      'Could not determine the output file associated with the transpiled input file'
+      'Could not determine the output file associated with the transpiled input file',
+  INSPECTOR_NOT_AVAILABLE: 'Inspector protocol only works for node 8+'
```
DominicKramer: What do you think about using `The V8 Inspector protocol is only available in Node 8+` instead?

bin/run-test.sh
```diff
@@ -32,6 +32,14 @@ function run {
 # Run test/coverage
 run build/test
```
DominicKramer: When we run the test here, we should make sure `GCLOUD_USE_INSPECTOR` is not set.  That is, the `GCLOUD_USE_INSPECTOR` environment variable could be set globally to true for someone that is testing the protocol.

Then when this script is run, the tests are actually being run with `GCLOUD_USE_INSPECTOR=true` twice.

We should also make sure that whatever changes we make in the script the script don't leak outside the script.  That is, if `GCLOUD_USE_INSPECTOR` has been set globally to `false`, we need to make sure it's value hasn't changed to `true` after this script is run.

bin/run-test.sh
```diff
@@ -32,6 +32,14 @@ function run {
...
 
+node_version_major=$(node -p "require('semver').major(process.version)")
+if test $node_version_major -ge 8
+then
+  echo 'Test with v8 Inspector'
```
DominicKramer: nit: `V8 Inspector`

test/test-v8debugapi.ts
```diff
@@ -110,6 +110,37 @@ function validateBreakpoint(breakpoint: stackdriver.Breakpoint): void {
...
   }
 }
+describe('v8debugapi', function() {
+  const config: DebugAgentConfig = extend({}, defaultConfig, {
+    workingDirectory: __dirname,
+    forceNewAgent_: true
+  });
+  const logger = new common.logger({ levelLevel: config.logLevel } as any as LoggerOptions);
+
+  it('Verify the correct debugapi was constructed', (done) => {
```
DominicKramer: The name `should use the correct debugapi` would match the naming convention in the file.

DominicKramer: Can this be moved in the existing `describe` block so we don't have an extra `describe` block if not needed.  If not, we should change the name of the `describe` block to be more specific.

test/test-v8debugapi.ts
```diff
@@ -110,6 +110,37 @@ function validateBreakpoint(breakpoint: stackdriver.Breakpoint): void {
...
+      const mapFiles = fileStats.selectFiles(/.map$/, process.cwd());
+      SourceMapper.create(mapFiles, function (err, mapper) {
+        assert(!err);
+
+        // TODO: Handle the case when mapper is undefined.
+        // TODO: Handle the case when v8debugapi.create returns null
+        api = debugapi.create(logger, config, jsStats, mapper as SourceMapper.SourceMapper) as DebugApi;
+        if (process.env.GCLOUD_USE_INSPECTOR && semver.satisfies(process.version, '>=8')) {
+          const inspectorapi = require('../src/agent/v8/inspector-debugapi');
+          assert.ok(api instanceof inspectorapi.InspectorDebugApi);
```
DominicKramer: We should also have tests to verify whether or not a message is displayed to the user with regards to supporting the inspector protocol in the 4 cases with inspector requested or not and the node version supporting inspector or not.

test_windows.js
```diff
@@ -0,0 +1,19 @@
+const spawn = require('child_process').spawn;
```
DominicKramer: I think it would be better to put this in the `bin` directory with the other scripts used for testing.

Also, for completeness we should add a license to the top of the file.

test_windows.js
```diff
@@ -0,0 +1,19 @@
...
+var path = require('path');
+var semver = require('semver');
+
+process.env.GCLOUD_PROJECT = 0
+process.env.CLOUD_DEBUG_ASSERTIONS = 1
+spawn(path.join('node_modules', '.bin', 'mocha.cmd'),
```
DominicKramer: Like in the Unix case, we should make sure an unset value of `GCLOUD_USE_INSPECTOR` is being used here and that the script doesn't leak environment variable state.

test_windows.js
```diff
@@ -0,0 +1,19 @@
...
+  [ path.join('build', 'test'), '--timeout 4000', '--R' ], {
+  stdio : 'inherit'
+}).on('close', function() {
```
DominicKramer: I think listening for [exit](https://nodejs.org/api/child_process.html#child_process_event_exit) might be better here.  The function would have parameters for the exit code and signal.  For each `spawn`, we need to record the exit code and exit from the `test_windows.js` script with a non-zero exit code if either `spawn` fails so that Appveyor records the test as a failure.

test_windows.js
```diff
@@ -0,0 +1,19 @@
...
+  if (semver.satisfies(process.version, '>=8')) {
+    process.env.GCLOUD_USE_INSPECTOR = true;
+    console.log('Run test with Inspector');
+    spawn(path.join('node_modules', '.bin', 'mocha.cmd'),
+      [ path.join('build', 'test'), '--timeout 4000', '--R' ], {
+      stdio : 'inherit'
```
DominicKramer: I think the options can be used to set environment variables for the child process.  If so, setting `GCLOUD_USE_INSPECTOR` there would probably be better.

bin/run-test.sh
gaofanmichael: I will modify the test script to reflect this.
Before running the test, I will cache GCLOUD_USE_INSPECTOR
inspector_flag=GCLOUD_USE_INSPECTOR
GCLOUD_USE_INSPECTOR=false
and after run the test, set it back.
GCLOUD_USE_INSPECTOR=inspector_flag

src/agent/util/utils.ts
gaofanmichael: SGTM

test/test-v8debugapi.ts
gaofanmichael: Done.

gaofanmichael: the other describe block is 'v8debugapi' but the debugapi is created in beforeEach(). I would be not convenient to create another debug in it block.

test_windows.js
gaofanmichael: Done.

test_windows.js
gaofanmichael: Done.

bin/run-test.sh
gaofanmichael: Checking back, setting GCLOUD_USE_INSPECTOR won't affect environmental variable globally.

appveyor.yml
```diff
@@ -20,6 +20,8 @@ test_script:
   # run tests
   - npm run build
   - node_modules/.bin/mocha build/test --timeout 4000 --R
+  - SET GCLOUD_USE_INSPECTOR=1
```
gaofanmichael: I checked there is on UNSET command on appveyor.js.

test/test-v8debugapi.ts
```diff
@@ -110,6 +110,44 @@ function validateBreakpoint(breakpoint: stackdriver.Breakpoint): void {
...
 }
+describe('debugapi selection', function() {
+  const config: DebugAgentConfig = extend({}, defaultConfig, {
+    workingDirectory: __dirname,
+    forceNewAgent_: true
+  });
+  const logger = new common.logger({ levelLevel: config.logLevel } as any as LoggerOptions);
+  let warnIsCalled = false;
+  logger.warn = function(_s: string) {
+    warnIsCalled = true;
```
DominicKramer: We should make sure that `warn` was called and that the message supplied to `warn` was the expected message.

appveyor.yml
```diff
@@ -3,6 +3,7 @@ environment:
   matrix:
   # node.js
   - nodejs_version: "4"
+
```
DominicKramer: nit: Remove this line.

test/test-v8debugapi.ts
gaofanmichael: Done.

bin/run-test.sh
kjin: @gaofanmichael Just for the sake of safety, can you make this line `GCLOUD_USE_INSPECTOR= run build/test` to make sure GCLOUD_USE_INSPECTOR is unset

src/agent/v8/debugapi.ts
```diff
@@ -41,12 +41,9 @@ interface DebugApiConstructor {
 }
 
 let debugApiConstructor: DebugApiConstructor;
-const nodeVersion = /v(\d+\.\d+\.\d+)/.exec(process.version);
 
-if (!nodeVersion || nodeVersion.length < 2) {
-  const dummyapi = require('./dummy-debugapi');
-  debugApiConstructor = dummyapi.DummyDebugApi;
-} else if (semver.satisfies(nodeVersion[1], '>=8')) {
+if (semver.satisfies(process.version, '>=8') &&
```
kjin: Can you remove `./dummy-debugapi` if it's no longer being used?

gaofanmichael: Done.

bin/run-test.sh
DominicKramer: We need to either do `GCLOUD_USE_INSPECTOR=` to unset the variable or `unset GCLOUD_USE_INSPECTOR` to make sure the first test doesn't use inspector.

PR #347, opened 2017-10-24
Allow a single direct access point to package.json

DominicKramer:


src/debuggee.ts
```diff
@@ -75,9 +79,9 @@ export class Debuggee {
    *     that the user has a way of noticing.
    *     TODO(ofrobots): has this been renamed to `status` in the API?
    */
-  constructor(properties: DebuggeeProperties) {
+  constructor(properties: DebuggeeProperties, appInfo: AppInfo) {
```
gaofanmichael: can we make appInfo one property in DebuggerProperties?

src/debuggee.ts
```diff
@@ -34,6 +33,11 @@ export interface DebuggeeProperties {
   statusMessage?: StatusMessage;
 }
 
+export interface AppInfo {
```
kjin: `PackageInfo` seems more accurate to me, since `AppInfo` implies that the package.json file being read is that of the application's.

src/debuggee.ts
kjin: Can we set `properties.agentVersion` before calling this constructor, so we don't have to pass `appInfo` in as an argument?

src/debuggee.ts
DominicKramer: Done

src/debuggee.ts
DominicKramer: Good idea.  Done.

DominicKramer: Done

PR #351, opened 2017-11-02
Skip 32bit platform warning test after node 8.9.0

gaofanmichael:
As asnyc stack traces will not be turned on by default after node 8.9.0,
The warning message of 'async stack trace is not available on 32 bit
platform' no longer emitted. We will not emit this message after node
8.9.0.

### DIFF
--- a/test/test-debuglet.ts
+++ b/test/test-debuglet.ts
@@ -345,7 +345,8 @@ describe('Debuglet', function() {
         // TODO: Handle the case where debuglet.debuggee is undefined
         assert.equal((debuglet.debuggee_ as Debuggee).project, projectId);
         const arch = process.arch;
-        if (semver.satisfies(process.version, '>=8.5') &&
+        if (semver.satisfies(process.version, '>=8.5.0') &&
+            semver.satisfies(process.version, '<8.9.0') &&
             (arch === 'ia32' || arch === 'x86') &&
             process.env.GCLOUD_USE_INSPECTOR) {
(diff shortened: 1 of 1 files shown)


test/test-debuglet.ts
```diff
@@ -346,6 +346,7 @@ describe('Debuglet', function() {
         assert.equal((debuglet.debuggee_ as Debuggee).project, projectId);
         const arch = process.arch;
         if (semver.satisfies(process.version, '>=8.5') &&
+            semver.satisfies(process.version, '<8.9') &&
```
DominicKramer: nit: It wouldn't hurt to make these `8.5.0` and `8.9.0` to be precise.

src/debuggee.ts
```diff
@@ -34,10 +34,10 @@ export interface DebuggeeProperties {
 }
 
 export class Debuggee {
-  private uniquifier: string;
```
DominicKramer: I think having these changes in a separate PR would be better.

DominicKramer:
I think the phrase `As asnyc stack traces will not be turned on by default after node 8.9.0...` should be changed to `As asnyc stack traces will not be turned on by default starting with node 8.9.0` because async stack traces are not turned on by default in `8.9.0` but the original phrasing makes it sound like it is not until `8.9.0` that that change occurs.

DominicKramer:
Also the title `Skip 32bit platform warning test after node 8.9.0` would probably be better written as `Skip 32bit platform warning starting with node 8.9.0`.

PR #358, opened 2017-11-06
feat: Enable isReady for debugger for GCF

gaofanmichael:
This PR mainly serve for GCF. The introduced isReady will return a
promise, and to be resolved either
1) time since last breakpoint fetch is within a heuristic duration.
2) breakpoint update completed.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -113,6 +113,34 @@ const formatBreakpoints = function(
           .join('\n');
 };
 
+/**
+ * CachedPromise stores a promise. This promise is initially null and can be
+ * assigned to a new promise when refresh is called. This promise will be
+ * resolved by calling member function resolve.
+ */
+export class CachedPromise {
+  private promise: Promise<void>|null;
+  private promiseResolve: () => void;
+
+  constructor() {
+    this.clear();
+  }
+  refresh(): void {
+    this.promise = new Promise<void>((resolve) => {
+      this.promiseResolve = resolve;
+    });
+  }
+  get(): Promise<void>|null {
+    return this.promise;
+  }
+  resolve(): void {
+    this.promiseResolve();
+  }
+  clear(): void{
+    this.promise = null;
+  }
+}
+
 export class Debuglet extends EventEmitter {
   private debug: Debug;
   private v8debug: DebugApi|null;
@@ -125,16 +153,12 @@ export class Debuglet extends EventEmitter {
   // breakpointFetched was resolved, which means breakpoint update was
   // successful.
   private breakpointFetchedTimestamp: number;
-  // breakpointFetched is a promise only to be resolved after breakpoint fetch
-  // was successful. It is returned by isReady.
-  private breakpointFetched: Promise<void>|null;
-  // breakpointFetchedResolve is the resolve function for breakpointFetched.
-  private breakpointFetchedResolve: () => void;
-  // debuggeeRegistered is a promise only to be resolved after debuggee
+  // breakpointFetched is a CachedPromise only to be resolved after breakpoint
+  // fetch was successful. Its stored promise will be returned by isReady().
+  private breakpointFetched: CachedPromise;
+  // debuggeeRegistered is a CachedPromise only to be resolved after debuggee
   // registration was successful.
-  private debuggeeRegistered: Promise<void>;
-  // debuggeeRegisteredResolve is the resolve function for debuggeeRegistered.
-  private debuggeeRegisteredResolve: () => void;
+  private debuggeeRegistered: CachedPromise;
 
   // Exposed for testing
   config: DebugAgentConfig;
@@ -199,9 +223,9 @@ export class Debuglet extends EventEmitter {
 
     this.breakpointFetchedTimestamp = -Infinity;
 
-    this.debuggeeRegistered = new Promise<void>((resolve) => {
-      this.debuggeeRegisteredResolve = resolve;
-    });
+    this.debuggeeRegistered = new CachedPromise();
+    this.breakpointFetched = new CachedPromise();
+    this.debuggeeRegistered.refresh();
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
@@ -360,18 +384,15 @@ export class Debuglet extends EventEmitter {
             PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
       return Promise.resolve();
     } else {
-      if (this.breakpointFetched) {
-        return this.breakpointFetched;
+      if (this.breakpointFetched.get() !== null) {
+        return this.breakpointFetched.get() as Promise<void>;
       }
-
-      this.breakpointFetched = new Promise<void>((resolve) => {
-        this.breakpointFetchedResolve = resolve;
-      });
-      this.debuggeeRegistered.then(() => {
+      this.breakpointFetched.refresh();
+      (this.debuggeeRegistered.get() as Promise<void>).then(() => {
         this.scheduleBreakpointFetch_(
           0 /*immediately*/, true /*only fetch once*/);
       });
-      return this.breakpointFetched;
+      return this.breakpointFetched.get() as Promise<void>;
     }
   }
 
@@ -562,7 +583,7 @@ export class Debuglet extends EventEmitter {
             // TODO: Handle the case when `result` is undefined.
             that.emit(
                 'registered', (result as {debuggee: Debuggee}).debuggee.id);
-            that.debuggeeRegisteredResolve();
+            that.debuggeeRegistered.resolve();
             if (!that.fetcherActive) {
               that.scheduleBreakpointFetch_(0, false);
             }
@@ -651,8 +672,8 @@ export class Debuglet extends EventEmitter {
                 }
                 that.breakpointFetchedTimestamp = Date.now();
                 if (once) {
-                  that.breakpointFetchedResolve();
-                  that.breakpointFetched = null;
+                  that.breakpointFetched.resolve();
+                  that.breakpointFetched.clear();
                 } else {
                   that.scheduleBreakpointFetch_(
                       that.config.breakpointUpdateIntervalSec, once);
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -114,21 +114,16 @@ const formatBreakpoints = function(
 };
 
 /**
- * CachedPromise stores a promise. This promise is initially null and can be
- * assigned to a new promise when refresh is called. This promise will be
- * resolved by calling member function resolve.
+ * CachedPromise stores a promise. This promise can be resolved by calling
+ * function resolve() and can only be resolved once.
  */
 export class CachedPromise {
-  private promise: Promise<void>|null = null;
-  private promiseResolve: (() => void)|null = null;
+  private promiseResolve: (() => void) | null;
+  private promise: Promise<void> = new Promise<void>((resolve) => {
+    this.promiseResolve = resolve;
+  });
 
-  refresh(): Promise<void> {
-    this.promise = new Promise<void>((resolve) => {
-      this.promiseResolve = resolve;
-    });
-    return this.promise;
-  }
-  get(): Promise<void>|null {
+  get(): Promise<void> {
     return this.promise;
   }
   resolve(): void {
@@ -138,22 +133,16 @@ export class CachedPromise {
       this.promiseResolve = null;
     }
   }
-  clear(): void {
-    this.promise = null;
-  }
 }
 export interface IsReady {
   isReady(): Promise<void>;
 }
 
 /**
- * IsReadyManager is a wrapper class to call debuglet.isReady().
+ * IsReadyManager is a wrapper class to use debuglet.isReady().
  */
 class IsReadyImpl implements IsReady {
-  private debuglet: Debuglet;
-  constructor(debuglet: Debuglet) {
-    this.debuglet = debuglet;
-  }
+  constructor(private debuglet: Debuglet) {}
   isReady(): Promise<void> {
     return this.debuglet.isReady();
   }
@@ -173,7 +162,7 @@ export class Debuglet extends EventEmitter {
   private breakpointFetchedTimestamp: number;
   // breakpointFetched is a CachedPromise only to be resolved after breakpoint
   // fetch was successful. Its stored promise will be returned by isReady().
-  private breakpointFetched: CachedPromise;
+  private breakpointFetched: CachedPromise|null;
   // debuggeeRegistered is a CachedPromise only to be resolved after debuggee
   // registration was successful.
   private debuggeeRegistered: CachedPromise;
@@ -244,8 +233,6 @@ export class Debuglet extends EventEmitter {
     this.breakpointFetchedTimestamp = -Infinity;
 
     this.debuggeeRegistered = new CachedPromise();
-    this.breakpointFetched = new CachedPromise();
-    this.debuggeeRegistered.refresh();
   }
 
   static normalizeConfig_(config: DebugAgentConfig): DebugAgentConfig {
@@ -404,16 +391,13 @@ export class Debuglet extends EventEmitter {
             PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
       return Promise.resolve();
     } else {
-      const result = this.breakpointFetched.get();
-      if (result) {
-        return result;
-      }
-      this.breakpointFetched.refresh();
-      (this.debuggeeRegistered.get() as Promise<void>).then(() => {
+      if (this.breakpointFetched) return this.breakpointFetched.get();
+      this.breakpointFetched = new CachedPromise();
+      this.debuggeeRegistered.get().then(() => {
         this.scheduleBreakpointFetch_(
           0 /*immediately*/, true /*only fetch once*/);
       });
-      return this.breakpointFetched.get() as Promise<void>;
+      return this.breakpointFetched.get();
     }
   }
 
@@ -642,9 +626,13 @@ export class Debuglet extends EventEmitter {
               // We back-off from fetching breakpoints, and try to register
               // again after a while. Successful registration will restart the
               // breakpoint fetcher.
-              that.debuggeeRegistered.refresh();
-              that.breakpointFetched.resolve();
-              that.breakpointFetched.clear();
+              that.debuggeeRegistered = new CachedPromise();
+              if (that.breakpointFetched) {
+                // breakpointFetched is resolved when failure occurs, therefore
+                // GCF will not hang forever to wait non-fetchable breakpoints.
+                that.breakpointFetched.resolve();
+                that.breakpointFetched = null;
+              }
               that.scheduleRegistration_(
                   that.config.internal.registerDelayOnFetcherErrorSec);
               return;
@@ -656,9 +644,14 @@ export class Debuglet extends EventEmitter {
                 // re-registration, which will re-active breakpoint fetching.
                 that.logger.info('\t404 Registration expired.');
                 that.fetcherActive = false;
-                that.debuggeeRegistered.refresh();
-                that.breakpointFetched.resolve();
-                that.breakpointFetched.clear();
+                that.debuggeeRegistered = new CachedPromise();
+                if (that.breakpointFetched) {
+                  // breakpointFetched is resolved when failure occurs,
+                  // therefore GCF will not hang forever to wait non-fetchable
+                  // breakpoints.
+                  that.breakpointFetched.resolve();
+                  that.breakpointFetched = null;
+                }
                 that.scheduleRegistration_(0 /*immediately*/);
                 return;
 
@@ -699,8 +692,10 @@ export class Debuglet extends EventEmitter {
                 }
                 that.breakpointFetchedTimestamp = Date.now();
                 if (once) {
-                  that.breakpointFetched.resolve();
-                  that.breakpointFetched.clear();
+                  if (that.breakpointFetched) {
+                    that.breakpointFetched.resolve();
+                    that.breakpointFetched = null;
+                  }
                 } else {
                   that.scheduleBreakpointFetch_(
                       that.config.breakpointUpdateIntervalSec, once);
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -135,6 +135,14 @@ export class CachedPromise {
     }
   }
 }
+
+/**
+ * IsReady will return a promise to user after user starting the debug agent.
+ * This promise will be resolved when
+ * 1. Time since last listBreakpoint was within a heuristic time.
+ * 2. listBreakpoint completed successfully.
+ * 3. Debuggee expired or failed, listBreakpoint cannot be completed.
+ */
 export interface IsReady {
   isReady(): Promise<void>;
 }
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -119,26 +119,29 @@ const formatBreakpoints = function(
  * resolved by calling member function resolve.
  */
 export class CachedPromise {
-  private promise: Promise<void>|null;
-  private promiseResolve: () => void;
+  private promise: Promise<void>|null = null;
+  private promiseResolve: (() => void)|null = null;
 
-  constructor() {
-    this.clear();
-  }
-  refresh(): void {
+  refresh(): Promise<void> {
     this.promise = new Promise<void>((resolve) => {
       this.promiseResolve = resolve;
     });
+    return this.promise;
   }
   get(): Promise<void>|null {
     return this.promise;
   }
   resolve(): void {
-    this.promiseResolve();
+    // Each promise can be resolved only once.
+    if (this.promiseResolve) {
+      this.promiseResolve();
+      this.promiseResolve = null;
+    }
   }
-  clear(): void{
+  clear(): void {
     this.promise = null;
   }
+
 }
 
 export class Debuglet extends EventEmitter {
@@ -384,8 +387,9 @@ export class Debuglet extends EventEmitter {
             PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
       return Promise.resolve();
     } else {
-      if (this.breakpointFetched.get() !== null) {
-        return this.breakpointFetched.get() as Promise<void>;
+      const result = this.breakpointFetched.get();
+      if (result) {
+        return result;
       }
       this.breakpointFetched.refresh();
       (this.debuggeeRegistered.get() as Promise<void>).then(() => {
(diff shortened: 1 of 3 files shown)


src/agent/debuglet.ts
```diff
@@ -322,6 +349,32 @@ export class Debuglet extends EventEmitter {
...
+        return this.breakpointFetched;
+      }
+      this.debuggeeRegistered.then(() => {
+        this.breakpointFetched = new Promise<void>((resolve) => {
+          this.breakpointFetchedResolve = resolve;
+        });
+        this.scheduleBreakpointFetch_(
+            0 /*immediately*/, true /*only fetch once*/);
+        return this.breakpointFetched;
+      });
```
DominicKramer: It looks like at this point the function doesn't return anything, which is automatically converted to a `Promise<void>` since the function is marked `async`.  However, the returned `Promise<void>` is not `this.breakpointFetched` and so the code won't work as expected.

src/index.ts
```diff
@@ -50,6 +50,5 @@ export function start(options: DebugAgentConfig|StackdriverConfig): Debuglet|
   debuglet = new Debuglet(debug, agentConfig);
   debuglet.start();
 
-  // We return the debuglet to facilitate testing.
-  return agentConfig.testMode_ ? debuglet : undefined;
+  return debuglet;
```
DominicKramer: Why did this line change?

test/test-debuglet.ts
```diff
@@ -924,6 +924,34 @@ describe('Debuglet', function() {
...
+                        .post(REGISTER_PATH)
+                        .reply(200, {debuggee: {id: DEBUGGEE_ID}})
+                        .get(BPS_PATH + '?successOnTimeout=true')
+                        .reply(200, {breakpoints: []});
+      debuglet.start();
+      const debugPromise = debuglet.isReady();
+      debugPromise
+          .then(() => {
+            setTimeout(function() {
+              debuglet.stop();
```
DominicKramer: I think you would want `asserts` here to verify breakpoints are fetched when the promise is resolved.

test/test-debuglet.ts
```diff
@@ -924,6 +924,34 @@ describe('Debuglet', function() {
       debuglet.start();
     });
 
+    it('should resolve promises before exiting user functions', function(done) {
```
DominicKramer: I don't think there is any relationship between the timing of the promise resolution and exiting a user function that we can make or that we need.  Instead we only need to ensure a breakpoint fetch (and necessary state updates) have occurred before the `Promise` is resolved.

src/agent/debuglet.ts
```diff
@@ -62,6 +62,12 @@ const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
     ' are CAPTURE and LOG.';
 
+// PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS is a heuristic duration that we set
+// to force the debug agent to return a new promise for isReady. The value is
+// the average of Stackdriver debugger hanging get duration (40s) and TCP
+// time-out on GCF (540s).
+const PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS = 290 * 1000;
```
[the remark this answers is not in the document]
kjin: nit^2: `(540 + 40) / 2 * 1000`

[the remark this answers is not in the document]
gaofanmichael: Done.

src/agent/debuglet.ts
```diff
@@ -114,6 +120,21 @@ export class Debuglet extends EventEmitter {
...
 
+  // breakpointFetchedTimestamp represents the last timestamp when
+  // breakpointFetched was resolved, which means breakpoint update was
+  // successful.
+  private breakpointFetchedTimestamp: number;
+  // breakpointFetched is a promise only to be resolved after breakpoint fetch
+  // was successful. It is returned by isReady.
+  private breakpointFetched: Promise<void>|null;
+  // breakpointFetchedResolve is the resolve function for breakpointFetched.
+  private breakpointFetchedResolve: (() => void);
```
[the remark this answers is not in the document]
gaofanmichael: Done.

src/agent/debuglet.ts
```diff
@@ -322,6 +349,32 @@ export class Debuglet extends EventEmitter {
   }
 
   /**
+   * isReady returns a promise that only resolved if the last breakpoint update
+   * happend within a duration (PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS). This
+   * feature is mainly used in Google Cloud Function (GCF), as it is a
+   * serverless environment and we wanted to make sure debug agent always
+   * captures the snapshots.
+   */
+  async isReady(): Promise<void> {
```
kjin: `async` is unnecessary if you don't use `await` in the function body.

src/agent/debuglet.ts
```diff
@@ -114,6 +120,21 @@ export class Debuglet extends EventEmitter {
...
+  // breakpointFetched is a promise only to be resolved after breakpoint fetch
+  // was successful. It is returned by isReady.
+  private breakpointFetched: Promise<void>|null;
+  // breakpointFetchedResolve is the resolve function for breakpointFetched.
+  private breakpointFetchedResolve: (() => void);
+  // debuggeeRegistered is a promise only to be resolved after debuggee
+  // registration was successful.
+  private debuggeeRegistered: Promise<void>;
+  // debuggeeRegisteredResolve is the resolve function for debuggeeRegistered.
+  private debuggeeRegisteredResolve: (() => void);
```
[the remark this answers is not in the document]
gaofanmichael: Done.

src/index.ts
gaofanmichael: because the user will call 
```javascript
const debug = require('@google/cloud-debugger').start();
const debugReady = debug.isReady();
```
we cannot return undefined in `start`

src/agent/debuglet.ts
```diff
@@ -522,15 +576,20 @@ export class Debuglet extends EventEmitter {
    * @param {number} seconds
    * @private
    */
-  scheduleBreakpointFetch_(seconds: number): void {
+  scheduleBreakpointFetch_(seconds: number, onlyOnce: boolean): void {
```
kjin: Add JSDoc for `onlyOnce`. Also, naming suggestion -- `once` is just as descriptive as `onlyOnce`.

src/agent/debuglet.ts
```diff
@@ -593,8 +651,14 @@ export class Debuglet extends EventEmitter {
                   that.logger.info(formatBreakpoints(
                       'Active Breakpoints: ', that.activeBreakpointMap));
                 }
-                that.scheduleBreakpointFetch_(
-                    that.config.breakpointUpdateIntervalSec);
+                if (onlyOnce) {
+                  that.breakpointFetchedTimestamp = Date.now();
```
kjin: Would you want to set the fetched timestamp regardless of the value of `onlyOnce`?

src/agent/debuglet.ts
```diff
@@ -322,6 +349,32 @@ export class Debuglet extends EventEmitter {
...
+  async isReady(): Promise<void> {
+    if (Date.now() < this.breakpointFetchedTimestamp +
+            PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      return Promise.resolve();
+    } else {
+      if (this.breakpointFetched) {
+        return this.breakpointFetched;
+      }
+      this.debuggeeRegistered.then(() => {
+        this.breakpointFetched = new Promise<void>((resolve) => {
```
kjin: This looks like `CachedPromise` in #341, since this Promise also goes stale. Is it possible to use that abstraction for `breakpointFetched` (and also `debuggeeRegistered`)? For the latter, since it never goes stale you could use a value of `Infinity` for time until stale.

test/test-debuglet.ts

test/test-debuglet.ts

src/agent/debuglet.ts
gaofanmichael: Removed.

src/agent/debuglet.ts
gaofanmichael: Checked back on this. I think separating listBreakpoint out would be tricky because it also calls scheduleBreakpointFetch for regular scheduling. Instead, I have made change in isReady to make sure the correct promise (breakpointFetched) is returned.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Yes. thanks for spotting this out. I think it makes sense to update the timestamp every time updateBreakpoint is called.

src/agent/debuglet.ts
gaofanmichael: So in #341, `CachedPromise` would also include a time variable. I have thought to use that abstraction here as well as we have two promises (`breakpointFetched`, `debuggeeRegistered`). But checking back their behaviors, `debuggeeRegistered` will be resolved after debuggee has been successfully registered and that's done. However, for `breakpointFetched`, when it has been resolved, it needs to be set to null so that next time when GCF calls `isReady', it can create a new promise instead of using the stale one. The logics would be more clean if not using abstraction in this case.

src/agent/debuglet.ts
```diff
@@ -129,12 +129,12 @@ export class Debuglet extends EventEmitter {
...
   private breakpointFetched: Promise<void>|null;
   // breakpointFetchedResolve is the resolve function for breakpointFetched.
-  private breakpointFetchedResolve: (() => void);
+  private breakpointFetchedResolve: () => void;
   // debuggeeRegistered is a promise only to be resolved after debuggee
   // registration was successful.
   private debuggeeRegistered: Promise<void>;
   // debuggeeRegisteredResolve is the resolve function for debuggeeRegistered.
-  private debuggeeRegisteredResolve: (() => void);
+  private debuggeeRegisteredResolve: () => void;
```
kjin: Nit - Given the similarity of `breakpointFetched` and `debuggeeRegistered` I would still like to see sort of abstraction, especially because you already wrote one in another PR.

test/test-debuglet.ts
```diff
@@ -924,33 +924,41 @@ describe('Debuglet', function() {
...
-      debuglet.start();
-      const debugPromise = debuglet.isReady();
-      debugPromise
-          .then(() => {
+        const scope = nock(API)
+                          .post(REGISTER_PATH)
+                          .reply(200, {debuggee: {id: DEBUGGEE_ID}})
+                          .get(BPS_PATH + '?successOnTimeout=true')
+                          .reply(200, {breakpoints: [breakpoint]})
+                          .get(BPS_PATH + '?successOnTimeout=true')
```
kjin: Nit: `.get(...).twice().reply(...)`

src/agent/debuglet.ts
gaofanmichael: Done.

test/test-debuglet.ts

src/agent/debuglet.ts
```diff
@@ -107,6 +113,34 @@ const formatBreakpoints = function(
...
+/**
+ * CachedPromise stores a promise. This promise is initially null and can be
+ * assigned to a new promise when refresh is called. This promise will be
+ * resolved by calling member function resolve.
+ */
+export class CachedPromise {
+  private promise: Promise<void>|null;
+  private promiseResolve: () => void;
+
+  constructor() {
```
DominicKramer: nit: You could just have `private promise: Promise<void>|null = null` and then not have a constructor at all.

src/agent/debuglet.ts
```diff
@@ -318,11 +369,34 @@ export class Debuglet extends EventEmitter {
...
+   * feature is mainly used in Google Cloud Function (GCF), as it is a
+   * serverless environment and we wanted to make sure debug agent always
+   * captures the snapshots.
+   */
+  isReady(): Promise<void> {
+    if (Date.now() < this.breakpointFetchedTimestamp +
+            PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      return Promise.resolve();
+    } else {
+      if (this.breakpointFetched.get() !== null) {
```
DominicKramer: If you said something like:
```
const result = this.breakpointFetched.get();
if (result) {
  return result;
}
```
then you shouldn't need the cast.

If you do need it, prefer `this.debuggeeRegistered.get()!` over `this.debuggeeRegistered.get() as Promise<void>`.  The first syntax makes it clear that you are saying the result is not null.  The second is more vague and could mean that the type is different from `Promise<void>`, but treat it as a `Promise<void>`.

test/test-debuglet.ts
```diff
@@ -77,6 +77,41 @@ function verifyBreakpointRejection(re: RegExp, body: {breakpoint: any}) {
...
+    cachedPromise.refresh();
+    (cachedPromise.get() as Promise<void>).then(() => {
+      done();
+    });
+    cachedPromise.resolve();
+  });
+  it('CachedPromise.get() will be null if it has been cleared', (done) => {
+    this.timeout(2000);
+    let cachedPromise = new CachedPromise();
+    assert(cachedPromise.get() === null);
```
DominicKramer: I would use `assert.stringEqual(cachedPromise.get(), null)` here.  That way if the assert fails, the actual value will be printed with the expected value.  Otherwise it will just print that `false` was encountered but `true` was expected.

src/agent/debuglet.ts
```diff
@@ -113,6 +113,34 @@ const formatBreakpoints = function(
...
+
+  constructor() {
+    this.clear();
+  }
+  refresh(): void {
+    this.promise = new Promise<void>((resolve) => {
+      this.promiseResolve = resolve;
+    });
+  }
+  get(): Promise<void>|null {
```
kjin: I recommend lazily initializing `this.promise` here if it's null rather than have a separate `refresh()` function for it. This way, you can ensure the return type of `get()` is non-null.

src/index.ts
ofrobots: The type definition for `start` should be updated. Secondly, let's return an opaque object with an `isReady` function rather than exposing the full `debuglet`.

src/agent/debuglet.ts
```diff
@@ -318,11 +369,34 @@ export class Debuglet extends EventEmitter {
...
+            PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      return Promise.resolve();
+    } else {
+      if (this.breakpointFetched.get() !== null) {
+        return this.breakpointFetched.get() as Promise<void>;
+      }
+      this.breakpointFetched.refresh();
+      (this.debuggeeRegistered.get() as Promise<void>).then(() => {
+        this.scheduleBreakpointFetch_(
+          0 /*immediately*/, true /*only fetch once*/);
```
ofrobots: The `boolean` argument seems like a code smell. It suggests to me that we should refactor `scheduleBreakpointFetch` into two functions. One that does with the 'scheduling bits' and the other that deals with the 'fetching' bits. This usage here only needs to deal with the latter.

src/agent/debuglet.ts
```diff
@@ -318,11 +369,34 @@ export class Debuglet extends EventEmitter {
...
+   */
+  isReady(): Promise<void> {
+    if (Date.now() < this.breakpointFetchedTimestamp +
+            PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      return Promise.resolve();
+    } else {
+      if (this.breakpointFetched.get() !== null) {
+        return this.breakpointFetched.get() as Promise<void>;
+      }
+      this.breakpointFetched.refresh();
```
ofrobots: This code here suggests that it would be nice to have `refresh` return the new promise.

src/agent/debuglet.ts
```diff
@@ -115,6 +149,17 @@ export class Debuglet extends EventEmitter {
...
+  // breakpointFetchedTimestamp represents the last timestamp when
+  // breakpointFetched was resolved, which means breakpoint update was
+  // successful.
+  private breakpointFetchedTimestamp: number;
+  // breakpointFetched is a CachedPromise only to be resolved after breakpoint
+  // fetch was successful. Its stored promise will be returned by isReady().
+  private breakpointFetched: CachedPromise;
+  // debuggeeRegistered is a CachedPromise only to be resolved after debuggee
+  // registration was successful.
+  private debuggeeRegistered: CachedPromise;
```
ofrobots: The only call to `debuggeeRegistered.refesh()` I see is in the constructor. This suggests to me that there is a single promise we use forever. This implies that `debuggeeRegistered.get().then(foo)` will run `foo` immediately (after a microtask hop).

This doesn't seem correct to me.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Added to your suggested pattern and removed cast.

test/test-debuglet.ts

src/agent/debuglet.ts
gaofanmichael: I did this because sometime we don't want to refresh the promise in get. For example, if previous `breakpointFetch` promised not being resolved yet, we wanted to return that promise by calling `get()` instead of refreshing a new one.

src/agent/debuglet.ts
gaofanmichael: there are two processes that will call `scheduleBreakpointFetch`, one is by the underlying scheduling from debuggee registration, the other is scheduled by `isReady()`. Here when the 'fetching' bits fails or has errors, it will also re call `scheduleBreakpointFetch` to reschedule. Therefore we cannot fully remove the Boolean variable to differentiate whether the scheduling is from debuggee or `isReady()`

src/agent/debuglet.ts
gaofanmichael: Done.

gaofanmichael: `refresh()` just refreshed the `CachedPromise.promise` so that when calling `scheduleBreakpointFetch` afterwards, that promise can be resolved. we won't resolve staled promised here. However, we will return promise after `scheduleBreakpointFetch` so will need to call `get` in the end. I guess refresh can return the newly created promise but not necessary.

src/agent/debuglet.ts
gaofanmichael: Thanks for pointing out! This is a bug and I just have it fixed with the new commit. PTAL.

src/index.ts
gaofanmichael: Done. Added an IsReadyManager class to only expose isReady.

gaofanmichael:
@ofrobots @DominicKramer I have comments addressed. PTAL :)

src/agent/debuglet.ts
```diff
@@ -107,6 +113,49 @@ const formatBreakpoints = function(
...
+  }
+  clear(): void {
+    this.promise = null;
+  }
+}
+
+/**
+ * IsReadyManager is a wrapper class to call debuglet.isReady().
+ */
+export class IsReadyManager {
```
DominicKramer: I would recommend using an `IsReady` interface here and have an `IsReadyImpl` (or another name if you prefer) that implements the interface.  Then the places that are currently referencing this class, can instead reference the interface since the class details are not actually needed.  This allows the implementation to be hidden.

That is, use:
```
export interface IsReady {
  isReady(): Promise<void>;
}
```
and 
```
class IsReadyImpl implements IsReady {
  private debuglet: Debuglet;
  constructor(debuglet: Debuglet) {
    this.debuglet = debuglet;
  }

  isReady(): Promise<void> {
    return this.debuglet.isReady();
  }
}
```

gaofanmichael: I agree. Thanks for bringing it out. I have it updated.

src/agent/debuglet.ts
```diff
@@ -107,6 +113,52 @@ const formatBreakpoints = function(
...
+  }
+}
+export interface IsReady {
+  isReady(): Promise<void>;
+}
+
+/**
+ * IsReadyManager is a wrapper class to call debuglet.isReady().
+ */
+class IsReadyImpl implements IsReady {
```
DominicKramer: I think our style guide would suggest writing this as:
```typescript
class IsReadyImpl implements IsReady {
  constructor(private debuglet: Debuglet) {}
  isReady(): Promise<void> {
    return this.debuglet.isReady();
  }
}
```

src/agent/debuglet.ts
```diff
@@ -318,11 +389,35 @@ export class Debuglet extends EventEmitter {
...
+    if (Date.now() < this.breakpointFetchedTimestamp +
+            PROMISE_RESOLVE_CUT_OFF_IN_MILLISECONDS) {
+      return Promise.resolve();
+    } else {
+      const result = this.breakpointFetched.get();
+      if (result) {
+        return result;
+      }
+      this.breakpointFetched.refresh();
+      (this.debuggeeRegistered.get() as Promise<void>).then(() => {
```
DominicKramer: If you want to state that `this.debuggeeRegistered.get()` is not `null`, prefer the syntax `this.debuggeeRegistered.get()!.then(...` instead of `(this.debuggeeRegistered.get() as Promise<void>).then(...`.

src/agent/debuglet.ts
```diff
@@ -541,18 +642,23 @@ export class Debuglet extends EventEmitter {
...
-
             // TODO: Address the case where `response` is `undefined`.
             switch ((response as http.ServerResponse).statusCode) {
               case 404:
                 // Registration expired. Deactivate the fetcher and queue
                 // re-registration, which will re-active breakpoint fetching.
                 that.logger.info('\t404 Registration expired.');
                 that.fetcherActive = false;
+                that.debuggeeRegistered.refresh();
+                that.breakpointFetched.resolve();
```
DominicKramer: Why is `breackpointFetched` resolved here?

src/agent/debuglet.ts
```diff
@@ -541,18 +642,23 @@ export class Debuglet extends EventEmitter {
               // We back-off from fetching breakpoints, and try to register
               // again after a while. Successful registration will restart the
               // breakpoint fetcher.
+              that.debuggeeRegistered.refresh();
+              that.breakpointFetched.resolve();
```
DominicKramer: Why is `breakpointFetched` resolved here?

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Per discussion, we do not want GCF hang forever when failure occurs. I added comments in code.

src/agent/debuglet.ts
```diff
@@ -107,6 +113,41 @@ const formatBreakpoints = function(
           .join('\n');
 };
 
+/**
+ * CachedPromise stores a promise. This promise can be resolved by calling
+ * function resolve() and can only be resolved once.
+ */
+export class CachedPromise {
+  private promiseResolve: (() => void) | null;
```
DominicKramer: nit: Use `private promiseResolve: (() => void) | null = null` so that the initial value of `promiseResolve` is `null`.

src/agent/debuglet.ts
```diff
@@ -107,6 +113,41 @@ const formatBreakpoints = function(
...
+  private promise: Promise<void> = new Promise<void>((resolve) => {
+    this.promiseResolve = resolve;
+  });
+
+  get(): Promise<void> {
+    return this.promise;
+  }
```
DominicKramer: nit: Add newline.

src/agent/debuglet.ts
```diff
@@ -175,6 +229,10 @@ export class Debuglet extends EventEmitter {
 
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap = {};
+
+    this.breakpointFetchedTimestamp = -Infinity;
+
```
DominicKramer: nit: Add `this.breakpointFetched = null` to explicitly show that we want the value to be initially `null`.  Otherwise, it could look like we forgot to initialize the variable.

src/agent/debuglet.ts
```diff
@@ -541,18 +626,32 @@ export class Debuglet extends EventEmitter {
               // We back-off from fetching breakpoints, and try to register
               // again after a while. Successful registration will restart the
               // breakpoint fetcher.
+              that.debuggeeRegistered = new CachedPromise();
```
DominicKramer: What if `this.debuggeeRegistered` is non-null here before being set to `new CachedPromise()`.  Could it be possible that the previous value of `this.debuggeeRegistered` is never resolved?

src/agent/debuglet.ts
```diff
@@ -541,18 +626,32 @@ export class Debuglet extends EventEmitter {
...
             }
-
             // TODO: Address the case where `response` is `undefined`.
             switch ((response as http.ServerResponse).statusCode) {
               case 404:
                 // Registration expired. Deactivate the fetcher and queue
                 // re-registration, which will re-active breakpoint fetching.
                 that.logger.info('\t404 Registration expired.');
                 that.fetcherActive = false;
+                that.debuggeeRegistered = new CachedPromise();
```
DominicKramer: What if `this.debuggeeRegistered` is non-null here before being set to `new CachedPromise()`.  Could it be possible that the previous value of `this.debuggeeRegistered` is never resolved?

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: Done.

src/agent/debuglet.ts
gaofanmichael: we will resolve breakpointFetched below -- that is the promise wait on debuggeeRegistered.

src/agent/debuglet.ts
```diff
@@ -107,6 +113,50 @@ const formatBreakpoints = function(
...
+    if (this.promiseResolve) {
+      this.promiseResolve();
+      this.promiseResolve = null;
+    }
+  }
+}
+
+/**
+ * IsReady will return a promise to user after user starting the debug agent.
+ * This promise will be resolved when
```
DominicKramer: nit: Specify `when one of the following is true:`

src/agent/debuglet.ts
```diff
@@ -175,6 +238,12 @@ export class Debuglet extends EventEmitter {
 
     /** @private {Object.<string, Boolean>} */
     this.completedBreakpointMap = {};
+
+    this.breakpointFetched = null;
+
```
DominicKramer: nit: Remove this newline.

src/agent/debuglet.ts
```diff
@@ -175,6 +238,12 @@ export class Debuglet extends EventEmitter {
...
+    this.breakpointFetchedTimestamp = -Infinity;
+
```
DominicKramer: nit: Remove this newline.

src/agent/debuglet.ts
```diff
@@ -591,15 +688,37 @@ export class Debuglet extends EventEmitter {
...
     }, seconds * 1000).unref();
   }
 
   /**
+   * updatePromise_ is called when debuggee is expired. debuggeeRegistered
+   * CachedPromise will be refreshed. Also, breakpointFetched CachedPromise will
+   * be resolved so that uses (such as GCF users) will not hang forever to wait
+   * non-fetchable breakpoints.
+   */
+  updatePromise_() {
```
ofrobots: drop the trailing underscore. Should this be private?

PR #359, opened 2017-11-10
test: add installation verification

DominicKramer:


test/test-install.ts
```diff
@@ -0,0 +1,97 @@
+
```
ofrobots: copyright

test/test-install.ts
```diff
@@ -0,0 +1,97 @@
...
+  afterEach(async function() {
+    this.timeout(TIMEOUT_MS);
+    if (installDir) {
+      await rimrafP(installDir);
+    }
+  });
+
+  TS_CODE.forEach((code) => {
+    it(`Should install and work with the Typescript code:\n${code}`, async function() {
```
ofrobots: nit: s/Should/should

test/test-install.ts
```diff
@@ -0,0 +1,97 @@
...
+      }, log);
+      await spawnP('node', [INDEX_JS], {
+        cwd: installDir,
+        stdio
+      }, log);
+    });
+  });
+
+  JS_CODE.forEach((code) => {
+    it(`Should install and work with the Javascript code:\n${code}`, async function() {
```
ofrobots: nit: s/Should/should

test/utils.ts
```diff
@@ -0,0 +1,66 @@
+
```
ofrobots: Copyright

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "system-test/**/*.ts"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "test/fixtures"
   ]
 }
```
ofrobots: Add newline while you're here.

package.json
```diff
@@ -78,7 +86,10 @@
     "compile-scripts": "tsc -p scripts-tsconfig.json",
     "fix": "gts fix",
     "pretest": "npm run compile",
-    "posttest": "npm run check"
+    "posttest": "npm run check",
+    "prepack": "npm run compile",
+    "pretest-install": "npm run compile",
+    "test-install": "mocha build/test/test-install.js"
```
ofrobots: IIUC, `test-install` gets run either when we do `npm run test` or we do `npm run test-install`. Why do we need a special run script for it when test covers it?

Secondly, do you think it might be too expensive to include in the default test suite?

The client library team is actually testing the samples include in their repo. Have you looked at that? This smells a lot like a 'sample' that we should test to ensure it continues to work. Thoughts?

test/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+export const writeFileP: (path: string, data: string, encoding?: string) => Promise<void> = pify(writeFile);
+export const statP: (path: string) => Promise<Stats> = pify(stat);
+export const tmpDirP: () => Promise<string> = pify(tmp.dir);
+export const rimrafP: (f: string) => Promise<void> = pify(rimraf);
+
+export function nodule(nodule: string) {
+  return path.relative(BUILD_DIRECTORY, `node_modules/${nodule}`);
+}
+
+export function existsP(path: string): Promise<boolean> {
```
gaofanmichael: I think it might be more helpful to change the name of the function to something more descriptive like expanding the meaning of 'P' as some of the function name P mean 'path' where others mean 'process'.

test/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (code === 0) {
+        exit();
+      } else {
+        exit(new Error(`Process ${childProcess.pid} exited with code ${code}.`));
+      }
+    });
+  });
+}
+
+export function spawnP(command: string, args?: string[], options?: SpawnOptions, log?: (text: string) => void): Promise<void> {
```
gaofanmichael: This just reminds me something --  can we also config the gts to fix test as well as src. In that way we don't have a line too long.

test/test-install.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const TS_CODE_1 = `import * as debug from '@google-cloud/debug-agent';`;
+const TS_CODE_2 = `import * as debug from '@google-cloud/debug-agent';
+debug.start();
+`;
+const TS_CODE_3 = `import * as debug from '@google-cloud/debug-agent';
+debug.start({ allowExpressions: true });
+`;
+
+const TS_CODE = [TS_CODE_1, TS_CODE_2, TS_CODE_3];
```
gaofanmichael: TS_CODE_ARRAY?

test/test-install.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const JS_CODE_1 = `require('@google-cloud/debug-agent').start()`;
+const JS_CODE_2 = `require('@google-cloud/debug-agent').start({})`;
+const JS_CODE_3 = `require('@google-cloud/debug-agent').start({
+  allowExpressions: true
+})`;
+
+const JS_CODE = [JS_CODE_1, JS_CODE_2, JS_CODE_3];
```
gaofanmichael: JS_CODE_ARRAY?

test/utils.ts
DominicKramer: The P in this case refers to Promise.  This is the same code used by the trace agent, but we could use different names if it makes it easier to reason about the code.

package.json
DominicKramer: I've removed the `test-install` since it is not needed.

I don't think these are too expensive to include in the default test suite.  Also, they are different from samples in that they test basic functionality, existence of types, etc. as apposed to displaying how to use the agent.

test/test-install.ts
DominicKramer: Added

test/test-install.ts
DominicKramer: Fixed

test/test-install.ts
DominicKramer: Fixed

test/utils.ts
DominicKramer: Added

tsconfig.json
DominicKramer: Added

test/test-install.ts
DominicKramer: Updated

test/test-install.ts
DominicKramer: Updated

DominicKramer:
@ofrobots @gaofanmichael PTAL

test/test-install.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  afterEach(async function() {
+    this.timeout(TIMEOUT_MS);
+    if (installDir) {
+      await rimrafP(installDir);
+    }
+  });
+
+  TS_CODE_ARRAY.forEach((code) => {
+    it(`should install and work with the Typescript code:\n${code}`, async function() {
```
kjin: Nit: I'd rather see a mid-level describe which describes the code than a multiline `it` statement. Maybe you can map elements in `[TS|JS]_CODE_ARRAY` to high-level descriptions of the code snippets?

test/utils.ts
kjin: I prefer the `-P` suffix given that all pify'd functions already exist purely for convenience. Having to type out `Promise` every time in place of `P` is cumbersome.

I remember replicating this pattern from another codebase... not sure where it started.

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "system-test/**/*.ts"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "test/fixtures"
```
kjin: Is this line necessary?

DominicKramer: Not anymore.  I originally had test code in the fixtures directory, but its not there now.

test/test-install.ts
DominicKramer: Fixed

PR #368, opened 2017-11-29
chore: Update Readme to include isReady interface

gaofanmichael:


README.md
```diff
@@ -189,6 +189,23 @@ process.env['GCLOUD_USE_INSPECTOR'] = true;
 require('@google-cloud/debug-agent').start({ ... });
 ```
 
+The stackdriver Debugger also introduces a new `isReady` method that returns a `Promise` that is resolved in either of the three scenarios.
```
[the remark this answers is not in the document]
kjin: Nit: Uppercase stackdriver.

README.md
```diff
@@ -189,6 +189,23 @@ process.env['GCLOUD_USE_INSPECTOR'] = true;
...
+1. The debug agent has received snapshot information from the Stackdriver service.
+2. The last snapshot received from Stackdriver service is relatively recent.
+3. The debug agent has determined that it is not possible to receive snapshot information.
+
+This is needed in environments such as Google Cloud Functions where, without this functionality, application code is completed before the debug agent has received its snapshot information.
+
+```js
```
kjin: It would be nice to show a before/after for user code on GCF (with annotations on why the "before" doesn't work).

README.md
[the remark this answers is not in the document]
gaofanmichael: Done.

README.md
gaofanmichael: "before code" was quite similar to quick start, example or debugger agent settings sections. I have added additional explanation for the reason why "before code" won't work.

kjin: The updated explanation still doesn't inform readers on why snapshot information is never updated. Instead of stating when the function is needed, can you phrase the explanation around when to use it instead?

kjin: I think the additional elaboration adds too much detail concerning the use of the Stackdriver Debugger. How about this:

"In order for Stackdriver Debugger to work on GCF, users should call the `isReady` function and wait for the returned `Promise` to resolve before exiting the cloud function. The purpose of this is to allow the Debug Agent enough time to sync reasonably up-to-date snapshot information; in most cases this is instantaneous and should not significantly affect peak latency." (Can you confirm that the last statement is true?)

gaofanmichael: Mostly right. Only occasionally it can delay up to 40 seconds when the snapshot was not updated for a while and the next polling is going to happen in 40 seconds.

kjin: Got it. Thanks!

PR #369, opened 2017-12-06
chore: Run `gts fix` and manually fix some errors

DominicKramer:


src/agent/v8/legacy-debugapi.ts
```diff
@@ -194,26 +194,22 @@ export class V8DebugApi implements debugapi.DebugApi {
...
     const listener =
-        this.onBreakpointHit.bind(this, breakpoint, function(err: Error) {
+        this.onBreakpointHit.bind(this, breakpoint, (err: Error) => {
           const currTime = Date.now();
           if (currTime > timesliceEnd) {
             logsThisSecond = 0;
             timesliceEnd = currTime + 1000;
           }
           print(
-              // TODO: Address the case where `breakpoint.logMessageFormat` is
```
jinwoo: Is this TODO no longer relevant? It seems we're still ignoring that case.

test/test-max-data-size.ts
```diff
@@ -75,32 +73,26 @@ describe('maxDataSize', function() {
...
+    api.set(bp, (err1) => {
       assert.ifError(err1);
-      // TODO: Determine how to remove this cast to any.
-      api.wait(bp, function(err2?: Error) {
+      api.wait(bp, (err2?: Error) => {
         assert.ifError(err2);
-        // TODO: Determine how to remove this cast to any.
-        assert(bp.variableTable.some(function(v) {
-          // TODO: Handle the case when v is undefined
-          // TODO: Handle the case when v.status is undefined
```
jinwoo: Similar question about these TODOs.

src/agent/debuglet.ts
```diff
@@ -276,7 +275,7 @@ export class Debuglet extends EventEmitter {
   async start(): Promise<void> {
     const that = this;
     process.on('warning', (warning) => {
-      if ((warning as any).code ===
+      if ((warning as {} as {code: string}).code ===
```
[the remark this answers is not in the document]
DominicKramer: Fixed.  However, I do not know why `Error` doesn't have a `code` attribute already so that this cast is not needed.  That investigation can be done in another PR.

src/agent/v8/legacy-debugapi.ts
DominicKramer: Good catch.  I've added the comment back.

test/test-max-data-size.ts
DominicKramer: For tests, if the value of a variable is `undefined`, accessing an attribute on that variable will cause an Error that will fail the test.  Thus, I think it is safe to use the syntax `someVar!.someAttribute` in tests.

src/agent/debuglet.ts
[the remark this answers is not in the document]
jinwoo: You should use `NodeJS.ErrnoException` for errors that have the `code` field.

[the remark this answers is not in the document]
DominicKramer: Fixed

Issue #367, opened 2017-11-22
Wrong types


agmontpetit:
I'm using Node with TypeScript and I'm facing an issue with types when starting the debugger as suggested in the [Quick Start](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs#quick-start) section:
```
import debugAgent = require('@google-cloud/debug-agent');
debugAgent.start({ allowExpressions: true });
```

It seems the type definition is wrong and more params (other than `allowExpressions`) are required:
```
Argument of type '{ allowExpressions: true; }' is not assignable to parameter of type 'DebugAgentConfig | StackdriverConfig'.
```

DominicKramer:
Thank you for letting us know about this.  This issue is being addressed in PR #370.

agmontpetit:
Thanks for the fix, @DominicKramer!

DominicKramer:
@google-cloud/debug-agent version 2.3.1 addresses this issue.

DominicKramer:
@agmontpetit Please let me know if you still have any issues using version 2.3.1.  Thanks.

agmontpetit:
@DominicKramer It works perfectly! Thanks!

DominicKramer:
No problem.  I'm glad to help.

PR #370, opened 2017-12-06
fix: Address compilation errors when using the agent with Typescript

DominicKramer:
fixes: #367


src/agent/config.ts
```diff
@@ -119,54 +119,86 @@ export interface DebugAgentConfig extends common.AuthenticationConfig {
...
-  forceNewAgent_: boolean;
+  forceNewAgent_?: boolean;
 
   /**
    * Uses by tests to cause the start() function to return the debuglet.
    */
+  testMode_?: boolean;
+}
+
+export interface ResolvedDebugAgentConfig extends DebugAgentConfig {
```
kjin: Do [`Partial<T>`](https://www.typescriptlang.org/docs/handbook/advanced-types.html) types work here?

kjin:
Additional nit -- PR title should be `fix:`, not `bug:`, right?

src/agent/config.ts
DominicKramer: Nice.  I wasn't aware of `Partial<T>` types.  I've updated the code to use them.

src/agent/config.ts
```diff
@@ -16,7 +16,11 @@
 
 import * as common from '../types/common';
 
-export interface DebugAgentConfig extends common.AuthenticationConfig {
+export type DebugAgentConfig = {
+  [K in keyof ResolvedDebugAgentConfig]?: Partial<ResolvedDebugAgentConfig[K]>
```
kjin: Does `export type DebugAgentConfig = Partial<ResolvedDebugAgentConfig>` work?

Also, can you check to see that doc comments still show up in VS Code intellisense?

DominicKramer: `export type DebugAgentConfig = Partial<ResolvedDebugAgentConfig>` does work.  Also, VS Code intellisense still shows the docs for `DebugAgentConfig`.

DominicKramer:
@kjin I uploaded a new commit that added additional installation tests.

Issue #366, opened 2017-11-22
Gracefully handle unexpected source maps


montmanu:
When encountering a source map that the agent doesn't expect (e.g., a css source map), the agent seem to fail to initialize. Here's a sample log entry:

```
ERROR:@google-cloud/debug-agent: Error processing the sourcemaps. Error: An error occurred while processing the sourcemap filesError: No sources listed in the sourcemap file public/vi-assets/main-b7c14129f872a062abe21d8edb364c0d.css.map
```

DominicKramer:
Thanks for letting us know about this issue.  It is being addressed in PR #371 .

DominicKramer:
`@google-cloud/debug-agent` version `2.3.2` addresses this issue.

montmanu:
thank you!

PR #371, opened 2017-12-07
fix: index only .js.map source map files

DominicKramer:
fixes: #366


ofrobots:
Can you add a test for this?

src/agent/debuglet.ts
```diff
@@ -266,6 +266,21 @@ export class Debuglet extends EventEmitter {
     return extend(true, {}, defaultConfig, config, envConfig);
   }
 
+  static async findFiles(shouldHash: boolean, baseDir: string): Promise<{
```
ofrobots: nit: not very readable; a named type would be better. Is it needed though? Doesn't the compiler infer these from the return type?

DominicKramer: I've added a named type.  The compiler can infer the return type, but I need to reference the type when I create a `findResults` variable.

PR #373, opened 2017-12-08
 fix: `process` event listeners are removed on stop

DominicKramer:
When the agent is started, a `warning` event listener is added
to `process`.  With this change, the listener is removed when
the agent is stopped.

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -244,11 +244,12 @@ export class Debuglet extends EventEmitter {
     this.breakpointFetchedTimestamp = -Infinity;
     this.debuggeeRegistered = new CachedPromise();
 
-    const debugAgentWarningListener = (warning: NodeJS.ErrnoException) => {
+    const that = this;
+    function debugAgentWarningListener(warning: NodeJS.ErrnoException) {
       if (warning.code === 'INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE') {
-        this.logger.info(utils.messages.ASYNC_TRACES_WARNING);
+        that.logger.info(utils.messages.ASYNC_TRACES_WARNING);
       }
-    };
+    }
 
     this.warningListener = debugAgentWarningListener;
   }
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -244,12 +244,11 @@ export class Debuglet extends EventEmitter {
     this.breakpointFetchedTimestamp = -Infinity;
     this.debuggeeRegistered = new CachedPromise();
 
-    const that = this;
-    function debugAgentWarningListener(warning: NodeJS.ErrnoException) {
+    const debugAgentWarningListener = (warning: NodeJS.ErrnoException) => {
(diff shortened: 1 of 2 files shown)


src/agent/debuglet.ts
```diff
@@ -164,6 +164,9 @@ export class Debuglet extends EventEmitter {
   private project: string|null;
   private controller: Controller;
   private completedBreakpointMap: {[key: string]: boolean};
+  private warningListener: ((warning: NodeJS.ErrnoException) => void)&{
+    fromGcpDebugAgent: boolean;
```
jinwoo: `fromGcpDebugAgent` is only for unit tests, right? Would it make more sense to make it an optional field then?

I'm not very happy with adding a debugging-only property here but I don't have any other good ideas for doing that.

DominicKramer: I can make it optional.  I also don't like this solution, but don't know of anything better to assist in tests.  I think just counting the number of listeners is too fragile.

DominicKramer: I've uploaded another solution.

src/agent/debuglet.ts
```diff
@@ -242,6 +243,15 @@ export class Debuglet extends EventEmitter {
     this.breakpointFetched = null;
     this.breakpointFetchedTimestamp = -Infinity;
     this.debuggeeRegistered = new CachedPromise();
+
+    const that = this;
```
jinwoo: It'll be better to use a fat arrow function and remove this `that = this` hack.

```typescript
const debugAgentWarningListener = (warning: ...) => {
  ...
};
```

But I'm not sure `fn.name` in the test is still `debugAgentWarningListener`.

DominicKramer: It still works with arrow functions.  I've updated the code.

jinwoo: Great. I like this better than the original one. BTW now that I look at the code again, why not just make `warningListener` or `debugAgentWarningListener` a private member function?

src/agent/debuglet.ts
```diff
@@ -274,11 +284,7 @@ export class Debuglet extends EventEmitter {
    */
   async start(): Promise<void> {
     const that = this;
```
ofrobots: unused?

src/agent/debuglet.ts
```diff
@@ -242,6 +243,15 @@ export class Debuglet extends EventEmitter {
     this.breakpointFetched = null;
     this.breakpointFetchedTimestamp = -Infinity;
     this.debuggeeRegistered = new CachedPromise();
+
+    const that = this;
+    function debugAgentWarningListener(warning: NodeJS.ErrnoException) {
```
ofrobots: Do we still need this? The issue has been fixed in Node at this point, and we will no longer be getting this warning, correct?

DominicKramer: Yes, I forgot about that.  I'm closing this in favor of PR #376

DominicKramer:
Closed in favor of PR #376.

Issue #377, opened 2017-12-12
GKE: Error scanning the filesystem


blerest:
When I start the debug agent on GCP GKE I have the following error:

> ERROR:@google-cloud/debug-agent: Error scanning the filesystem. Error: ENOENT: no such file or directory, lstat '/proc/15/fd/11'

I'm not able to see the application on debug UI.

```
require('@google-cloud/debug-agent').start({
  allowExpressions: true,
  serviceContext: {
    service: 'service-sms',
  },
});
```

My cluster has the the `cloud_debugger` OAuth scope.

ofrobots:
Thanks for the bug report. We'll take a look at this.

In the meanwhile, my theory is that you are building your docker image with the app (specifically `package.json`) being in the root directory. We try to scan the file system starting from the root directory to find all the `.js` files that could potentially see a breakpoint. In this case since your `package.json` is in the root directory, we probably ended up scanning the entire filesystem (we need to fix this).

As a work-around, can you try putting your app into a sub-directory, say `/app`? Let us know if that helps.

blerest:
> As a work-around, can you try putting your app into a sub-directory, say /app? Let us know if that helps.

Yeah I just changed the WORKDIR and it's now working.

DominicKramer:
This issue is being addressed by PR #379

DominicKramer:
PR #381 implements a change where the agent won't start if the working directory is a root directory to avoid scanning the entire filesystem during startup for Javascript files.

DominicKramer:
`@google-cloud/debug-agent` version `2.3.2` addresses this issue.

PR #379, opened 2017-12-14
fix: Address startup failure from file access errors

DominicKramer:
fixes: #377


test/fixtures/project-cannot-read-files/cannot-read-1.js
```diff
@@ -0,0 +1 @@
+This file cannot be read.
```
ofrobots: This might end up breaking certain kind of tooling.

Would it be possible for you to generate these files dynamically?

DominicKramer: In later commits, I've mocked out `fs` so that the `project-cannot-read-files` directory is no longer used at all.

src/agent/io/scanner.ts
```diff
@@ -223,29 +230,40 @@ function findFiles(baseDir: string, regex: RegExp): Promise<string[]> {
...
+          shasum.update(d);
+        }
+        lines++;
+      });
+      byLine.on('end', () => {
+        if (error) {
+          reject(error);
+        }
+        else {
+          let d: string|undefined;
```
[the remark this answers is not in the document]
DominicKramer: I've updated it to
```js
const hash = shouldHash ? shasum.digest('hex') : undefined;
resolve({hash, lines});
```
This is needed because
```js
shouldHash && 'some text'
```
is of type `string` if `shouldHash` is `true` and of type `boolean` if `shouldHash` is `false`.  However, `hash` is expected to be of type `string|undefined`.

src/agent/debuglet.ts
```diff
@@ -294,6 +294,9 @@ export class Debuglet extends EventEmitter {
     try {
       fileStats = await scanner.scan(
           !id, that.config.workingDirectory, /.js$|.map$/);
+      fileStats.errors().forEach(err => {
```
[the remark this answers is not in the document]
DominicKramer: Fixed

test/test-debuglet.ts
```diff
@@ -444,6 +447,73 @@ describe('Debuglet', () => {
       debuglet.start();
     });
 
+    it('should not fail if files cannot be read', (done) => {
+      const MOCKED_DIRECTORY = process.cwd();
+      const errors: Array < {
```
ofrobots: nit: not very readable. Extract into a named type.

PR #381, opened 2017-12-20
fix: refuse to start if working dir is root dir

DominicKramer:
fixes: #377

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -294,8 +294,13 @@ export class Debuglet extends EventEmitter {
     const workingDir = that.config.workingDirectory;
     // Don't continue if the working directory is a root directory
     if (path.join(workingDir, '..') === workingDir) {
-      that.logger.error(`Refusing to start with \`workingDirectory\` set to a root directory:  '${workingDir}'`);
-      that.emit('initError', new Error(`Cannot start the agent when the working directory is a root directory`));
+      that.logger.error(
+          `Refusing to start with \`workingDirectory\` set to a root directory:  '${
+              workingDir}'`);
+      that.emit(
+          'initError',
+          new Error(
+              `Cannot start the agent when the working directory is a root directory`));
       return;
     }
 
@@ -740,7 +745,6 @@ export class Debuglet extends EventEmitter {
           formatBreakpoints('Server breakpoints: ', updatedBreakpointMap));
     }
     breakpoints.forEach((breakpoint: stackdriver.Breakpoint) => {
-
       // TODO: Address the case when `breakpoint.id` is `undefined`.
       if (!that.completedBreakpointMap[breakpoint.id as string] &&
           !that.activeBreakpointMap[breakpoint.id as string]) {
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -291,24 +291,21 @@ export class Debuglet extends EventEmitter {
     const that = this;
     const stat = promisify(fs.stat);
 
-    const workingDir = that.config.workingDirectory;
-    // Don't continue if the working directory is a root directory
-    if (path.join(workingDir, '..') === workingDir) {
-      that.logger.error(
-          `Refusing to start with \`workingDirectory\` set to a root directory:  '${
-              workingDir}'`);
-      that.emit(
-          'initError',
-          new Error(
-              `Cannot start the agent when the working directory is a root directory`));
-      return;
-    }
-
     try {
       await stat(path.join(that.config.workingDirectory, 'package.json'));
     } catch (err) {
       that.logger.error('No package.json located in working directory.');
       that.emit('initError', new Error('No package.json found.'));
+      return;
+    }
+
+    const workingDir = that.config.workingDirectory;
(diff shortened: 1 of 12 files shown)


src/agent/debuglet.ts
```diff
@@ -291,6 +291,19 @@ export class Debuglet extends EventEmitter {
...
+    const workingDir = that.config.workingDirectory;
+    // Don't continue if the working directory is a root directory
+    if (path.join(workingDir, '..') === workingDir) {
+      that.logger.error(
+          `Refusing to start with \`workingDirectory\` set to a root directory:  '${
+              workingDir}'`);
+      that.emit(
+          'initError',
+          new Error(
+              `Cannot start the agent when the working directory is a root directory`));
```
ofrobots: 1. This doesn't tell the user why we don't want to start if we're in the root directory.
1. Please provide a way (config) for users to silence this check and force a start even in the root directory.
1. The check for package.json should happen before this check. That is a more severe problem.

src/agent/debuglet.ts
```diff
@@ -291,6 +291,19 @@ export class Debuglet extends EventEmitter {
     const that = this;
     const stat = promisify(fs.stat);
 
+    const workingDir = that.config.workingDirectory;
+    // Don't continue if the working directory is a root directory
+    if (path.join(workingDir, '..') === workingDir) {
+      that.logger.error(
+          `Refusing to start with \`workingDirectory\` set to a root directory:  '${
```
kjin: Why are there two slightly different error strings for the logger and event emission? Can they be the same?

test/test-debuglet.ts
```diff
@@ -377,6 +377,32 @@ describe('Debuglet', () => {
       assert.deepEqual(mergedConfig, compareConfig);
     });
 
+    it('should not start when workingDirectory is the a directory', (done) => {
```
kjin: nit: typo ("the root directory"?)

test/test-debuglet.ts
```diff
@@ -377,6 +377,32 @@ describe('Debuglet', () => {
...
+      const debug = new Debug({}, packageInfo);
+      const config = extend({}, defaultConfig, {workingDirectory: path.sep});
```
kjin: Does passing `path.sep` accurately represent Windows root directories?

test/test-debuglet.ts
```diff
@@ -377,6 +377,32 @@ describe('Debuglet', () => {
...
+      const debuglet = new Debuglet(debug, config);
+      let text = '';
+      debuglet.logger.error = (str: string) => {
+        text += str;
+      };
+
+      debuglet.on('initError', (err: Error) => {
+        assert.ok(err);
```
kjin: Maybe use `assert.ifError` here? And three lines below use `assert.ok`

test/test-debuglet.ts
DominicKramer: Yes, I've check that it works on Windows and OS X.

test/test-debuglet.ts
DominicKramer: `assert.ifError(err)` would cause the test to fail if `err` is defined.  Instead, I want to verify that `err` is defined and so I used `ok`.  I can switch the other `assert` to `assert.ok` since they are aliases if you think it is clearer.

test/test-debuglet.ts
DominicKramer: It is supposed to say "a root directory" because Windows can have multiple root directories.

src/agent/debuglet.ts
DominicKramer: I was trying to make the error message a little more concise than the log message, but I can have them the same.

test/test-debuglet.ts
DominicKramer: Updated `assert` to `assert.ok`.

test/test-debuglet.ts
DominicKramer: Fixed

src/agent/debuglet.ts
DominicKramer: I've updated the error message and log message to be the same.

src/agent/debuglet.ts
DominicKramer: Fixed 1 and 3.

DominicKramer: Fixed 2

src/agent/debuglet.ts
```diff
@@ -299,6 +299,20 @@ export class Debuglet extends EventEmitter {
       return;
     }
 
+    const workingDir = that.config.workingDirectory;
+    // Don't continue if the working directory is a root directory
+    // unless the user wants to force using the root directory
+    if (!that.config.allowRootAsWorkingDirectory &&
+        path.join(workingDir, '..') === workingDir) {
+      const message =
+          `Refusing to start the cloud debugger with \`workingDirectory\` set to a root ` +
```
ofrobots: nit: suggest text: `The working directory is a root directory. Disabling to avoid a scan of the entire filesystem for JavaScript files. Use config \`allowRootAsWorkingDirectory\` if you really want to do this.`

PR #384, opened 2018-01-05
chore: license check as posttest

jinwoo:
There are two problematic dependencies: log-driver & modelo. They come
from googleapis/nodejs-common. I created an issue for that:
https://github.com/googleapis/nodejs-common/issues/16. Until it's
resolved, this repo must have a package whitelist.

### DIFF
--- a/js-green-licenses.json
+++ b/js-green-licenses.json
@@ -1,0 +1,6 @@
+{
+  "packageWhitelist": [
+    "log-driver",
+    "modelo"
+  ]
+}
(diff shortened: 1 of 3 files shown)


js-green-licenses.json
```diff
@@ -0,0 +1,5 @@
+{
+  "packageWhitelist": [
+    "log-driver"
```
DominicKramer: You mentioned that Modelo is also a problem.  Why isn't it listed here?

jinwoo: Modelo was resolved over the weekend: https://github.com/kevinconway/Modelo.js/pull/20

jinwoo:
Filed https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/issues/387

PR #400, opened 2018-02-21
feat: GAE_DEPLOYMENT_ID used for minor version

DominicKramer:
The `GAE_DEPLOYMENT_ID` environment variable (with a fallback to
`GAE_MINOR_VERSION`) will be used as the default value of
`minorVersion_`.

Note that the `GAE_DEPLOYMENT_ID` environment
variable is now set on AppEngine Flex.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@google-cloud/common": {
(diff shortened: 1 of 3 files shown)


jinwoo:
Looks like tests must be updated?

DominicKramer:
PTAL

Issue #391, opened 2018-01-18
Transpiled Node.js apps that are sub-directories of the git repository don't work.


slavakl:
Similar to #51 but for transpiled apps.
Scenario:
Git repository with a directory serviceA/ that contains a Node.js app. i.e. 
serviceA/
--lib/
----utils.es6
--app.es6 

Repository is hosted in github. Source code transpiled with babel with source maps.
On a server(docker), it runs from /src folder, /src/serviceA.

When trying to set a breakpoint for app.es6, debugger says that corresponding source file cannot be found.
Breakpoint for utils.es6 "successfully" set on UI, but never hits.

Debugging through the agent shows that:

 - app.es6 cannot be found by findScriptsFuzzy. serviceA/app.es6 that comes from request doesn't match with /app.es6

 - serviceA/lib/utils.es6 cannot be found in source map's sources (sourcemapper.mappingInfo) as the relative path it's looking for is not resolved correctly.

It would be great if we could specify "rootFolder" that agent would use to search for sources and maps but also resolve all the relatives paths against, instead of using working directory. That would allow to use an agent with monorepos.

DominicKramer:
@slavakl Thank you for the feedback and sorry for the late reply.  We'll look into this and will let you know what we find.  In the meantime, have you tried the `appPathRelativeToRepository` option:
```
The path within your repository to the directory
containing the package.json for your deployed application. This should
be provided if your deployed application appears as a subdirectory of
your repository. Usually this is unnecessary, but may be useful in
cases where the debug agent is unable to resolve breakpoint locations
unambiguously.
```
It is used as follows:
```js
require('@google-cloud/debug-agent').start({
  appPathRelativeToRepository: 'some/path/'
});
```
It will be helpful to learn how this does or does not help with your problem so we can come up with a solution.  Thanks.

slavakl:
@DominicKramer thank you for the response. I'll try it later today. Seems that I overlooked this options. I tried to set wokingDir but it didn't help. Will report back whether appPathRelativeToRepository solves the problem.

slavakl:
I'm not sure if I set appPathRelativeToRepository right, but did't work in my case. Below is the log file plus some debug output that I added to the agent:

When running with only appPathRelativeToRepository set: 
```
require('@google-cloud/debug-agent').start({
   allowExpressions: true,
  appPathRelativeToRepository: 'serviceA',
  logLevel: 4,
});
[...]

PR #402, opened 2018-03-01
 fix: fix hitting breakpoints in transpiled subdirs

DominicKramer:
Fixes: #391


src/agent/io/sourcemapper.ts
```diff
@@ -43,6 +43,7 @@ export interface MapInfoOutput {
  * @private
  */
 function processSourcemap(
+    prefix: string|undefined,
```
ofrobots: is there benefit to allowing `undefined` here?

src/agent/io/sourcemapper.ts
```diff
@@ -108,9 +109,14 @@ function processSourcemap(
     }
 
     sources.forEach((src: string) => {
+      const key = path.normalize(path.join(prefix || '', src));
       infoMap.set(
-          path.normalize(src),
-          {outputFile: outputPath, mapFile: mapPath, mapConsumer: consumer});
+          key,
+          {
+            outputFile: path.normalize(path.join(prefix || '', outputPath)),
```
ofrobots: `path.join` normalizes as well, so no need to `normalize` ourselves.

Here and elsewhere.

src/agent/io/sourcemapper.ts
```diff
@@ -246,13 +259,14 @@ export class SourceMapper {
 }
 
 export function create(
+    prefix: string|undefined,
```
ofrobots: prefer `prefix` was never undefined.

I see that you needed to modify all the tests to pass in `undefined` by default. That isn't very readable. My preference would be to add an overload that accepts the prefix string and then uses type inspection, or argument count checks, to figure out whether a prefix was provided.

In the rest of this file, i.e. in the constructor, enforce that prefix is never undefined.

src/agent/io/sourcemapper.ts
DominicKramer: This can be `undefined` here since the value of `appPathRelativeToRepository` is eventually used as the value for `prefix` and `appPathRelativeToRepository` can be `undefined`.

In particular, I think having `undefined` to signify default behavior is easier to remember rather than having a special string value.  Specifically, the string that would cause default behavior is the empty string.

src/agent/io/sourcemapper.ts
DominicKramer: Oh, ok I didn't know that.  I'll update this.

ofrobots:
@DominicKramer what is the status of this one?

DominicKramer:
I'm investigating the feasibility of replacing this PR with PR #461 or something similar.  If we decide to go with PR #461, this PR will be closed and its ideas will be incorporated into #461.

PR #407, opened 2018-03-14
chore: update gcp-metadata to 0.6.3

DominicKramer:
`gcp-metadata` version `0.6.3` uses Typescript 2.7.  Therefore, for this module to compile, its Typescript version has been updated.

### DIFF
--- a/test/nocks.ts
+++ b/test/nocks.ts
@@ -44,9 +44,7 @@ export function register(validator?: Validator): nock.Scope {
   return nock('https://clouddebugger.googleapis.com')
       .post('/v2/controller/debuggees/register', validator)
       .once()
-      .reply(200, {
-        debuggee: {}
-      });
+      .reply(200, {debuggee: {}});
 }
 
 export function projectId(reply: string): nock.Scope {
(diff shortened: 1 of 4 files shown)


ofrobots:
This commit also changes the typescript version, so the issue title isn't accurate. Is the typescript change intentional?

DominicKramer:
The Typescript change is required by the gcp-metadata change.  I'll update the description.

test/nocks.ts
```diff
@@ -28,8 +28,8 @@ export type Validator = (body: {
 // TODO: Determine if the type of `validator` is correct.
 export function oauth2(validator?: Validator): nock.Scope {
   validator = validator || accept;
-  return nock('https://accounts.google.com')
-      .post('/o/oauth2/token', validator)
+  return nock(/https:\/\/(accounts\.google\.com|www\.googleapis\.com)/)
```
ofrobots: nit: With gcp-metadata@0.6.3 only one of these two urls is possible. You don't need to test either/or. I would suggest dropping the old version.

DominicKramer: Ok I'll drop the old version (https://accounts.google.com).

PR #425, opened 2018-05-07
 chore: add tests to verify evaluated expressions

DominicKramer:


test/test-evaluated-expressions.ts
```diff
@@ -0,0 +1,209 @@
...
+describe('debugger provides useful information', () => {
+  let api: debugapi.DebugApi;
+  const config =
+      extend({}, defaultConfig, {allowExpressions: true, forceNewAgent_: true});
+
+  before(done => {
+    // TODO: It appears `logLevel` is a typo and should be `level`.  However,
+    //       with this change, the tests fail.  Resolve this.
+    const logger =
+        new common.logger({levelLevel: config.logLevel} as {} as LoggerOptions);
```
kjin: `levelLevel` looks like a typo here?

test/test-evaluated-expressions.ts
```diff
@@ -0,0 +1,209 @@
...
+    scanner.scan(true, config.workingDirectory, /.js$/).then(fileStats => {
```
kjin: Nit: `/\.js$/` here and elsewhere

test/test-evaluated-expressions.ts
DominicKramer: Fixed

test/test-evaluated-expressions.ts
DominicKramer: Fixed

PR #431, opened 2018-05-16
chore: clean up test-e2e.ts

eyqs:
Update comments to be more correct and make lines 338 to 417 identical to lines 146 to 225, for consistency's sake.

### DIFF
--- a/system-test/test-e2e.ts
+++ b/system-test/test-e2e.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018 Google Inc. All Rights Reserved.
+ * Copyright 2015 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/system-test/test-e2e.ts
+++ b/system-test/test-e2e.ts
@@ -412,7 +412,6 @@ describe('@google-cloud/debug end-to-end behavior', () => {
 
           console.log('-- waiting before checking if the log was written');
           return Promise.all([breakpoint, delay(10 * 1000)]);
-          // TODO: Determine if the results parameter should be used.
         })
         .then((results: stackdriver.Breakpoint[]) => {
           // Check that the contents of the log is correct
--- a/system-test/test-e2e.ts
+++ b/system-test/test-e2e.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,7 +43,7 @@ interface Child {
   process?: cp.ChildProcess;
 }
 
-// This test could take up to 70 seconds.
+// This test could take up to 100 seconds.
 describe('@google-cloud/debug end-to-end behavior', () => {
   let api: Debugger;
 
@@ -162,7 +162,7 @@ describe('@google-cloud/debug end-to-end behavior', () => {
           });
           assert.ok(result, 'should find the debuggee we just registered');
           // TODO: Determine how to properly specify the signature of
-          // listDebuggees
+          // listBreakpoints
           return (api as {listBreakpoints: Function})
               .listBreakpoints(debuggeeId);
           // TODO: Determine if this type signature is correct.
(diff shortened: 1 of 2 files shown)


system-test/test-e2e.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
DominicKramer: For existing files, I don't think we should change the copyright date.

system-test/test-e2e.ts
```diff
@@ -43,7 +43,7 @@ interface Child {
   process?: cp.ChildProcess;
 }
 
-// This test could take up to 70 seconds.
+// This test could take up to 100 seconds.
```
DominicKramer: Was this failing on machine with 70 seconds because of a timeout, but was succeeding with 100 seconds?

system-test/test-e2e.ts
```diff
@@ -394,10 +396,12 @@ describe('@google-cloud/debug end-to-end behavior', () => {
             expressions: ['o'],
             log_message_format: 'o is: $0'
           });
+          // TODO: Determine if this type signature is correct.
         })
         .then((results: stackdriver.Breakpoint[]) => {
           // Check that the breakpoint was set, and then wait for the log to be
           // written to
+
```
DominicKramer: This looks like an unnecessary empty line.

system-test/test-e2e.ts
```diff
@@ -394,10 +396,12 @@ describe('@google-cloud/debug end-to-end behavior', () => {
             expressions: ['o'],
             log_message_format: 'o is: $0'
           });
+          // TODO: Determine if this type signature is correct.
```
DominicKramer: What is this TODO in reference to?

system-test/test-e2e.ts
```diff
@@ -408,6 +412,7 @@ describe('@google-cloud/debug end-to-end behavior', () => {
 
           console.log('-- waiting before checking if the log was written');
           return Promise.all([breakpoint, delay(10 * 1000)]);
+          // TODO: Determine if the results parameter should be used.
```
DominicKramer: What is this TODO in reference to?

test/debugger.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
DominicKramer: As with the e2e test file I don't think this date should change.

test/debugger.ts
```diff
@@ -122,8 +119,7 @@ export class Debugger extends common.ServiceObject {
       debuggeeId: string, options: {
         includeAllUsers?: boolean;
         includeInactive?: boolean;
-        stripResults?: boolean;
-        action?: string
+        action?: stackdriver.Action
```
kjin: Nit: I know you didn't change this line, but can you add `;` here?

system-test/test-e2e.ts
kjin: It looks like it's making the format more consistent with other sections...

system-test/test-e2e.ts
eyqs: The test always succeeds without timing out, but it ran for about 100 seconds, so I thought the comment was misleading. Do you think it should be removed entirely?

DominicKramer: In that case, I think the comment is good.

test/debugger.ts
```diff
@@ -109,10 +109,7 @@ export class Debugger extends common.ServiceObject {
    *     breakpoints set by all users, or just by the caller (default false).
    * @param {boolean=} options.includeInactive - Whether or not to include
    *     inactive breakpoints in the list (default false).
-   * @param {boolean=} options.stripResults - If set to true, breakpoints will be
```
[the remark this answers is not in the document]
DominicKramer: In that case, I think it would probably be better to remove this parameter in a separate PR and describe in the PR description that the param is being removed because it is deprecated.

system-test/test-e2e.ts
eyqs: I copied this in order to make lines 338 to 417 identical to lines 146 to 225. I guess it's in reference to `results` having type `Breakpoint[]`.

system-test/test-e2e.ts
eyqs: I copied this in order to make lines 338 to 417 identical to lines 146 to 225. I guess it's in reference to whether we should remove `results`, since nothing uses it.

Issue #417, opened 2018-04-19
Debugging transpiled code changes the breakpoint line number


xp500:
After applying https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/pull/402 locally, I managed to get the stackdriver debugger to work. The problem is that if I set a breakpoint on a typescript file the breakpoint line will change when the breakpoint hits

Before hitting
![image](https://user-images.githubusercontent.com/4530263/38971821-520798f4-4373-11e8-9c89-2c463892ecb6.png)
After hitting
![image](https://user-images.githubusercontent.com/4530263/38971838-5cb528ac-4373-11e8-8ad9-4197c736cd79.png)

I believe that the problem is on https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/src/agent/v8/legacy-debugapi.ts#L461 and https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/blob/master/src/agent/v8/inspector-debugapi.ts#L469 which changes the line number to the one set in the call frames. The problem is that the call frames refer to another file (the javascript file) and not the original typescript file where I set the breakpoint.

DominicKramer:
Thank you for letting us know about this and for looking into this.  As we work through this, we will keep this thread updated with status updates.

richardfickling:
This has an additional impact which is that breakpoints set on a line after compiled file ends won't work at all. For example, my `user.ts` file has 519 lines. Compiled as `user.js`, it has 434 lines. I can't add a breakpoint to `user.ts` after line 434

![image](https://user-images.githubusercontent.com/1130084/40577867-84f65290-60c0-11e8-962e-ff2e8e5aece3.png)

DominicKramer:
Thanks for letting us know about this.  Based on this feedback, I have updated the priority of this issue.

DominicKramer:
This issue has been fixed and released in version 2.6.0.  If you are still encountering problems, please let me know.  Thanks.

richardfickling:
@DominicKramer  The issue I reported above (unable to add breakpoints after end of original file) is still happening, although the original submitted issue (breakpoints moving) seems to be fixed.

PR #436, opened 2018-06-11
 fix: line numbers don't change in transpiled files

DominicKramer:
Fixes: #417

### DIFF
--- a/src/agent/v8/inspector-debugapi.ts
+++ b/src/agent/v8/inspector-debugapi.ts
@@ -466,7 +466,8 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       const captured = state.capture(
           callFrames, breakpoint, this.config, this.scriptMapper,
           this.v8Inspector);
-      if (breakpoint.location && breakpoint.location.path.toLowerCase().endsWith('.js')) {
+      if (breakpoint.location &&
+          breakpoint.location.path.toLowerCase().endsWith('.js')) {
         breakpoint.location.line = callFrames[0].location.lineNumber + 1;
       }
       breakpoint.stackFrames = captured.stackFrames;
--- a/src/agent/v8/inspector-debugapi.ts
+++ b/src/agent/v8/inspector-debugapi.ts
@@ -467,7 +467,7 @@ export class InspectorDebugApi implements debugapi.DebugApi {
           callFrames, breakpoint, this.config, this.scriptMapper,
           this.v8Inspector);
       if (breakpoint.location &&
-          path.extname(breakpoint.location.path).toLowerCase() === '.js') {
+          utils.isJavaScriptFile(breakpoint.location.path)) {
         breakpoint.location.line = callFrames[0].location.lineNumber + 1;
       }
       breakpoint.stackFrames = captured.stackFrames;
--- a/src/agent/v8/inspector-debugapi.ts
+++ b/src/agent/v8/inspector-debugapi.ts
@@ -467,7 +467,7 @@ export class InspectorDebugApi implements debugapi.DebugApi {
           callFrames, breakpoint, this.config, this.scriptMapper,
           this.v8Inspector);
       if (breakpoint.location &&
-          breakpoint.location.path.toLowerCase().endsWith('.js')) {
+          path.extname(breakpoint.location.path).toLowerCase() === '.js') {
         breakpoint.location.line = callFrames[0].location.lineNumber + 1;
       }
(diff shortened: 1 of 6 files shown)


DominicKramer:
PTAL

src/agent/v8/inspector-debugapi.ts
```diff
@@ -466,7 +466,8 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       const captured = state.capture(
           callFrames, breakpoint, this.config, this.scriptMapper,
           this.v8Inspector);
-      if (breakpoint.location) {
+      if (breakpoint.location &&
+          breakpoint.location.path.toLowerCase().endsWith('.js')) {
```
jinwoo: Maybe use `path.extname()`?

test/test-v8debugapi.ts
```diff
@@ -218,7 +221,8 @@ describeFn('debugapi selection on Node >=10', () => {
 describe('v8debugapi', () => {
   const config: ResolvedDebugAgentConfig = extend(
       {}, defaultConfig, {workingDirectory: __dirname, forceNewAgent_: true});
-  // TODO(dominickramer): It appears `logLevel` is a typo and should be `level`.  However,
+  // TODO(dominickramer): It appears `logLevel` is a typo and should be `level`.
+  // However,
   //       with this change, the tests fail.  Resolve this.
   const logger =
       new common.logger({levelLevel: config.logLevel} as {} as LoggerOptions);
```
jinwoo: Not this PR, but is `levelLevel` really a correct field name?

test/test-v8debugapi.ts
```diff
@@ -685,6 +690,30 @@ describe('v8debugapi', () => {
          });
        });
 
+    it.only(
```
jinwoo: Delete `only`.

DominicKramer: Good catch. Thanks.

test/test-v8debugapi.ts
DominicKramer: I think this should be addressed in a follow-up PR.

jinwoo: Sounds good. I'm just wondering how tests work when the log level is set incorrectly.

src/agent/v8/inspector-debugapi.ts
DominicKramer: Updated.

DominicKramer:
@jinwoo Both `.ts` and `.js` files are used because, to provide a nice user experience, breakpoints can be set in `.ts` files.  In this case, sourcemap information is used to set the breakpoint in the associated `.js` file.

src/agent/v8/inspector-debugapi.ts
```diff
@@ -466,7 +466,8 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       const captured = state.capture(
           callFrames, breakpoint, this.config, this.scriptMapper,
           this.v8Inspector);
-      if (breakpoint.location) {
+      if (breakpoint.location &&
+          path.extname(breakpoint.location.path).toLowerCase() === '.js') {
```
jinwoo: This seems like a good one-liner that can go to /util/utils.ts because it's used both here and in legacy-debugapi.ts. WDYT?

jinwoo:
@DominicKramer I understand that `.ts` files are used to provide a better user experience and I totally agree with that. What I don't understand is why/how stackdriver gives us both a `.ts` file path and a `.js` file path when a user sets a breakpoint.

test/test-v8debugapi.ts
```diff
@@ -685,6 +690,29 @@ describe('v8debugapi', () => {
          });
        });
 
+    it.only('should not change line number when breakpoints hit for transpiled files',
```
jinwoo: Another `only` ?

src/agent/v8/inspector-debugapi.ts
DominicKramer: Sounds good.

test/test-v8debugapi.ts
DominicKramer: I'm trying to determine why the tests are failing only on AppVeyor, and so I'm only running the one test in question.

src/agent/v8/inspector-debugapi.ts
DominicKramer: I've added the `utils.isJavaScriptFile` function.

PR #461, opened 2018-07-10
appPathResolver to allow arbitrary path mapping on behalf of agent

dylanscott:
Just thought I'd open a PR for this, per [my comment on this issue](https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/issues/391#issuecomment-403653747). No worries if you'd like to address this in a different way but this option would be very useful for me since the repository layout for some of my projects is more complicated than `appPathRelativeToRepository` allows for. Happy to iterate on naming/documentation of this flag


dylanscott:
I signed it!

DominicKramer:
Thank you for opening this PR.  I think it looks really good.  I'm looking into the use cases for using the `appPathRelativeToRepository` config to see how they interact with the `appPathResolver` config.

In particular, perhaps we could make `appPathResolver` to be of type `string | (path: string) => string`.

In the case when the type is a string, its value will be used similar to `appPathRelativeToRepository` (and then the `appPathRelativeToRepository` config wouldn't be needed).  This scenario could perhaps fit most use cases.

However, if more customization is needed, a function could specified for `appPathResolver` as your tests illustrate in this PR.

I'm still investigating whether this approach would work and would like to know what you think.  Thanks.

src/agent/config.ts
```diff
@@ -164,6 +164,58 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
...
+   *
+   * If the array `['/x/y/src/index.js', '/x/y/src/someDir/index.js']` or
+   * equivalently `undefined` is returned by the `pathResolver` function, the
+   * debug agent will not be able to set the breakpoint.
+   *
+   * If, however, the `pathResolver` function returned `['/x/y/src/index.js']`,
+   * for example, the debug agent would know to set the breakpoint in
+   * the `/x/y/src/index.js` file.
+   */
+  pathResolver?:
```
kjin: Is returning `undefined` semantically different from returning an empty array? If not, is there any reason to allow `undefined` as a return value?

src/agent/util/utils.ts
```diff
@@ -34,6 +37,52 @@ export interface Listener {
...
+    }
+
+    for (const path of calculatedPaths) {
+      if (knownFiles.indexOf(path) === -1) {
+        // tslint:disable-next-line:no-unused-expression
+        logger.warn(
+            `The 'pathResolver' config function returned a path '${path}' ` +
+            `that is not in the list of paths known to the debug agent ` +
+            JSON.stringify(knownFiles, null, 2)) +
+            ` only known paths can be returend. Continuing with the` +
```
kjin: `returned`

src/agent/util/utils.ts
```diff
@@ -34,6 +37,52 @@ export interface Listener {
...
+      return resolved;
+    }
+
+    for (const path of calculatedPaths) {
+      if (knownFiles.indexOf(path) === -1) {
+        // tslint:disable-next-line:no-unused-expression
+        logger.warn(
+            `The 'pathResolver' config function returned a path '${path}' ` +
+            `that is not in the list of paths known to the debug agent ` +
+            JSON.stringify(knownFiles, null, 2)) +
```
kjin: Move last `)` down two lines.

src/agent/util/utils.ts
```diff
@@ -34,6 +37,52 @@ export interface Listener {
...
+        return resolved;
+      }
+    }
+
+    return calculatedPaths;
+  }
+  return resolved;
+}
+
+export function _findScripts(
```
kjin: I would prefer we name this something else -- the leading underscore doesn't really describe why this function is different than `findScripts` other than that it is "more private".

src/agent/util/utils.ts
DominicKramer: `gts` placed that paren there.

kjin: I do not think that is the reason. The last two strings appear to be erroneously appended to the return result of `logger.warn(arg)` (rather than `arg` itself).

DominicKramer: Fixed

src/agent/util/utils.ts
DominicKramer: Fixed

src/agent/config.ts
DominicKramer: Returning `undefined` means "I don't want to perform a custom action, just used the resolved files that the debug agent think should be used".  The empty array means that the given path doesn't resolve to any files.

DominicKramer:
@kjin PTAL

src/agent/config.ts
kjin: Sounds good. I would suggest separately considering whether we can just allow the user to return a falsey value -- after all, I think we shouldn't ever treat `undefined` as an intentionally set value (I would prefer null) -- but I won't the block the PR on this.

PR #471, opened 2018-07-16
chore(deps): start using renovate

ofrobots:


JustinBeckwith:
Actually wait - then we should remove the Greenkeeper badge from the README right?

ofrobots:
Done. AFAICT there is no renovate badge?

JustinBeckwith:
Correct, no badge.

PR #477, opened 2018-07-18
test: use strictEqual in tests

JustinBeckwith:


test/test-debug-assert.ts
```diff
@@ -21,20 +21,20 @@ describe('debug-assert', () => {
...
 
     it('should not fire assertions when disabled', () => {
-      assert.equal(1, 2);
+      assert.strictEqual(1, 2);
     });
 
     it('should cover the full assert API', () => {
       Object.keys(realAssert).forEach((key) => {
-        realAssert.equal(
+        realassert.strictEqual(
```
DominicKramer: This looks like a typo.  It should be `realAssert`.

Issue #459, opened 2018-07-10
Why source-context.json instead of in code options?


kilianc:
The way I am using this today, I use $REVISION (git-hash) as `context.version`.

Reading the docs it looks like I need to generate, at container build time, a file with repo url and a commit hash.

Wouldn't it be easier to just pass the parameters to the agent at run-time? At this time I am injecting a new file at `docker build` time but it feels wrong and hacky.

DominicKramer:
Thanks for opening this issue.  I'm not sure, at the moment, if there is a better approach.  We will talk with the Stackdriver team to determine the best approach here and will keep this thread updated.  Thanks.

ofrobots:
I think this looks like a reasonable feature request. Right now our workflow is optimized towards the GAE use-cases which autogenerates these files – easy peasy. In other environments, the only mechanism we provide to the user is that they must generate the file manually. I am guessing @kilianc wants to do something like: (correct me if I am wrong)

IOW, I think you want to generate a file with this instead of generating a `source-context.json` file:

```ts
require('@google-cloud').start({
  sourceContext: { url: "https://github.com/me/cool-repo", revisionId: process.env.REVISION }
});
```

This looks reasonable to me!

PR #478, opened 2018-07-18
feat: accept SourceContext in config

ofrobots:
In preference to looking for a well-known source-contexts.json file, we
now accept sourceContext as part of the config as well.

Fixes: https://github.com/GoogleCloudPlatform/cloud-debug-nodejs/issues/459


kilianc:
<3 This is great!

PR #479, opened 2018-07-21
refactor: use console-log-level for logging

ofrobots:
Start using console-log-level so that we can drop the custom logging
code in @google-cloud/common. This also fixes the issue that the logs
were previously getting emitted to stdout instead of stderr.


src/agent/debuglet.ts
```diff
@@ -255,6 +256,20 @@ export class Debuglet extends EventEmitter {
...
+  //                     0        1        2       3       4        5
+  static LEVELNAMES = [ 'fatal', 'error', 'warn', 'info', 'debug', 'trace'];
+  static logLevelToName(level: number): string {
+    if (typeof level === 'string') {
+      level = parseInt(level, 10);
+    }
+    if (typeof level !== 'number') {
+      level = defaultConfig.logLevel;
+    }
+    if (level < 0) level = 0;
```
DominicKramer: For clarity, I think we should still use `{ ... }` around the body of the if statements.

src/agent/debuglet.ts
```diff
@@ -255,6 +256,20 @@ export class Debuglet extends EventEmitter {
...
+    if (level > 4) level = 4;
```
DominicKramer: Why isn't `else if` used here?

src/agent/debuglet.ts
ofrobots: Style guide is okay with this, and I like the improved readability from brevity.

src/agent/debuglet.ts
DominicKramer: Ok

ofrobots:
Related: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27506. Will bring that in once it lands.

src/agent/debuglet.ts
DominicKramer: I prefer always using brackets, but if the style guide is ok with this then so am I.

DominicKramer:
For record keeping, why did you decide to use `console-log-level`?

ofrobots:
console-log-level serves the purpose we need it for
* It can log to stderr – unlike log-driver.
* It has support for prefixes – something we need.
* It is not a global logger – unlike npmlog – so it is suitable for libraries to use.
* It is small, simple and well-maintained.

src/agent/debuglet.ts
```diff
@@ -14,38 +14,32 @@
  * limitations under the License.
  */
 
-import {AuthenticationConfig, Common} from '../types/common';
-const common: Common = require('@google-cloud/common');
-
+import * as assert from 'assert';
```
JustinBeckwith: Why import assert?

src/agent/debuglet.ts
```diff
@@ -14,38 +14,32 @@
...
 import * as debugapi from './v8/debugapi';
-
-import * as assert from 'assert';
-
-import * as stackdriver from '../types/stackdriver';
-import {DebugAgentConfig, ResolvedDebugAgentConfig} from './config';
-import {Debug, PackageInfo} from '../client/stackdriver/debug';
-import {Logger} from '../types/common';
 import {DebugApi} from './v8/debugapi';
+import consoleLogLevel = require('console-log-level');
```
JustinBeckwith: It looks like there is an @types/console-log-level package, can we use that?

ofrobots: Absolutely! That's why I wrote it.

src/agent/debuglet.ts
ofrobots: We have always imported `assert` – there are some debugging assertions in the code. `gts fix` reorganized the imports for whatever reason.

Issue #433, opened 2018-05-18
Update @google-cloud/common to the latest version


DominicKramer:

DominicKramer:
Work has begun on PR #448.

ofrobots:
PR #473 tracks this, so I am going to close this issue out.

PR #481, opened 2018-07-24
 chore: update @google-cloud/common to 0.20.3

DominicKramer:
Fixes: #433

### DIFF
--- a/test/auth-request.ts
+++ b/test/auth-request.ts
@@ -16,15 +16,18 @@
 
 import * as request from 'request';
 
-// TODO: Make the type of `options` more precise
-export const authRequest =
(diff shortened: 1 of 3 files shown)


test/auth-request.ts
```diff
@@ -16,15 +16,16 @@
...
-     (request.UrlOptions & request.CoreOptions),
-     callback: (err: Error, body: {}, response: request.RequestResponse) =>
-         void) => {
-      request(
-          options,
-          (err: Error, response: request.RequestResponse, body: {}) => {
-            callback(err, body, response);
-          });
-    };
+export async function authRequest(
```
ofrobots: Does this need to be an async function? this doesn't seem to use await anywhere in the body.

PR #486, opened 2018-07-30
chore: assert.deepEqual => assert.deepStrictEqual

kinwa91:
deepEqual is deprecated. Use deepStrictEqual instead


DominicKramer:
The failing tests are being investigated by issue #487.

Issue #446, opened 2018-06-20
Fix flaky test


JustinBeckwith:
The test `should re-fetch breakpoints on error` was flaking, so I had to skip it.
https://travis-ci.org/GoogleCloudPlatform/cloud-debug-nodejs/builds

Lets fix the flake and get it re-enabled :)

ofrobots:
Merging #466 into this issue. @DominicKramer had this comment over there:

> Flakiness in the unit tests were causing issues with landing PR #463. Fix the tests.

I continue to see flakiness in ~50% of the PRs.

ofrobots:
Another flake on https://ci.appveyor.com/project/ofrobots/cloud-debug-nodejs/build/1.0.1223 for #477.

ofrobots:
This is an important issue, but it is not a P1. This is being actively investigated by. Moving this to P2.

soldair:
so. another problem.
the most expected problem ever. the nocks in one test are being satisfied by requests from another.

we call. debuglet.stop() but debuglet occasionally has an inflight request that hits the nocks for the next test. All of the nocks are satisfied for the test. the test passed. It just consumes a mock setup for the next test.

this could be fixed one or all of a few ways. 

have debugglet.stop return a promise. resolving when there are no active fetches queued for registration or breakpoint fetch
tests handling nocks should be refactored as so
```
scope.done()
debuglet.stop().then(done)
```
:point_up: :sparkles: 

clearing all the nocks. this seems to work though i haven't finished walking the code path in nock.
it seems to clear both interceptors and pending requests.
```
afterEach(()=>{
  nock.clearAll()
})
```
:point_up: :man_shrugging: 

unique the test url so each gets a unique nock scope and pass in api url somehow
the code uses an api url constant and there is no api exposed atm to override this config.
```
let testUrl = API+Date.now()
let scope = nock(testUrl)
```
:point_up: :fearful: 

use mockable timers.
because this uses setTimeout these tests are much longer than they should.
... eh at some point we should just tighten the class up so it doesn't depend on so much
:point_up: :running_man:

soldair:
im going with the allow api url to be configured and its working really well.

here is a case where the net connect error comes from the first test but doesnt trigger until a few tests later.
 
```
      ✓ should resolve breakpointFetched promise when registration expires (95ms)
Some name Failed to re-register debuggee fake-project: Error: Debuglet not running
[...]

PR #493, opened 2018-08-03
chore: remove greenkeeper badge

JustinBeckwith:


JustinBeckwith:
@DominicKramer looks like we're still flakin' :(

DominicKramer:
@JustinBeckwith the flaky tests are being tracked by issue #446 that I am actively investigating.

DominicKramer:
The tests are failing only because of flakiness independent of this PR.  That issue is being tracked by issue #446.

That is, if I run the tests, Node version X will pass and Y will fail because of mocks not satisfied.  When I re-run the tests, Node version X will fail for the same reason why Y previously failed, and version Y will pass.  As a result, I am going to land this PR.

PR #505, opened 2018-08-17
test: use unique api url per nock

soldair:
fixes GoogleCloudPlatform/cloud-debug-nodejs#446
i hope =)

this adds the ability to configure the apiUrl used by
src/agent/controller.ts passing through src/agent/debuglet.ts

because debuglet makes or may still have requests open after stop()
 is called. It occasionally triggers nocks for other tests. 
nock is only global so we play its own game by making sure urls from 
one test have no way of matching another.

note: it seems easier to mock out debuglet.controller instead
of depending on nocks.

### DIFF
--- a/src/agent/config.ts
+++ b/src/agent/config.ts
@@ -259,7 +259,7 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
   testMode_: boolean;
 
   /**
-   * used by tests that nock to isolate nocks to their own scopes.
+   * used to set a default api url
    */
   apiUrl?: string;
 }
--- a/src/agent/config.ts
+++ b/src/agent/config.ts
@@ -257,6 +257,11 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
    * Uses by tests to cause the start() function to return the debuglet.
    */
   testMode_: boolean;
+
+  /**
+   * used by tests that nock to isolate nocks to their own scopes.
+   */
+  apiUrl?: string;
 }
 
 export interface StackdriverConfig extends GoogleAuthOptions {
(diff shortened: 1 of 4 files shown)


test/test-debuglet.ts
```diff
@@ -1333,3 +1362,17 @@ describe('Debuglet', () => {
...
+
+// a counter for unique test urls.
+let apiUrlInc = 0;
+
+/**
+ * returns a new config object to be passed to debuglet. always has apiUrl
+ * @param conf custom config values
+ */
+function debugletConfig(conf?: {}): (ResolvedDebugAgentConfig & {apiUrl:string}) {
+  const apiUrl = API + (++apiUrlInc);
```
ofrobots: nit: Move `API` to be a local variable here. Better, it would be nice if we didn't duplicate the constant here and it was exported from `controller.ts`. That doesn't have to be done as part of this change though.

soldair: i would like to see all api urls/defaults come from config.ts or a consistent place instead of the implementation. 

good call moving it into the function for now it'll make it harder just to use API in a new test instead of the unique name,

ofrobots: sgtm

ofrobots:
@soldair LMK when this is ready for another look.

soldair:
@ofrobots this is good for another look. perhaps @DominicKramer can help me as you're out this week.

src/agent/config.ts
```diff
@@ -257,6 +257,11 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
    * Uses by tests to cause the start() function to return the debuglet.
    */
   testMode_: boolean;
+
+  /**
+   * used by tests that nock to isolate nocks to their own scopes.
+   */
+  apiUrl?: string;
```
DominicKramer: Since this is only used by tests we should name it `apiUrl_` to indicate that its private.

soldair: this is a really great nit. ill fix it up

this change was largely aspirational because right now the api url is defined as a constant in the controller class. I really think it would be better if all of the defaults come from config with a convenient api for setting values.

PR #508, opened 2018-08-27
chore: add a synthfile, and make the build consistent

JustinBeckwith:


JustinBeckwith:
@DominicKramer I can't really figure out why the unit tests wouldn't be passing.  They fail for me locally with this setup as well.  I must be missing something 🤷‍♂️

kinwa91:
Perhaps a environment variable set within travis? (I have no access)

Issue #464, opened 2018-07-11
Unable to add breakpoints after end of original file


DominicKramer:
Issue #417 consisted of two problems.  PR #436 was created to fix all of issue #417, but the problem listed below is still and issue.

> This has an additional impact which is that breakpoints set on a line after compiled file ends won't work at all. For example, my user.ts file has 519 lines. Compiled as user.js, it has 434 lines. I can't add a breakpoint to user.ts after line 434

![image](https://user-images.githubusercontent.com/14253877/42588924-19281294-84f4-11e8-9841-b8e6fd669887.png)

PR #513, opened 2018-08-28
fix: allow snapshots in shorter transpiled files

DominicKramer:
If a source file had `n` lines and was transpiled to a JavaScript
file with `m` lines with `m < n`, and one tried to set a snapshot
on a line between `m+1` and `n` in the original source file,
setting the snapshot would incorrectly fail with an error stating
that the requested line was beyond the size of the file.

Fixes: #464


src/agent/v8/legacy-debugapi.ts
```diff
@@ -320,6 +307,19 @@ export class V8DebugApi implements debugapi.DebugApi {
       column += debugapi.MODULE_WRAP_PREFIX_LENGTH - 1;
     }
 
+    // TODO: Address the case where `fileStats[matchingScript]` is `null`.
+    if (line >= (this.fileStats[matchingScript] as FileStats).lines) {
+      return utils.setErrorStatusAndCallback(
+          cb, breakpoint, StatusMessage.BREAKPOINT_SOURCE_LOCATION,
+          utils.messages.INVALID_LINE_NUMBER + matchingScript + ':' +
+              (breakpoint.location as stackdriver.SourceLocation).line +
```
soldair: same as the other. should this be `line`?

src/agent/v8/inspector-debugapi.ts
```diff
@@ -339,6 +324,20 @@ export class InspectorDebugApi implements debugapi.DebugApi {
       column += debugapi.MODULE_WRAP_PREFIX_LENGTH - 1;
     }
 
+    // TODO: Address the case where `breakpoint.location` is `null`.
+    // TODO: Address the case where `fileStats[matchingScript]` is `null`.
+    if (line >= (this.fileStats[matchingScript] as FileStats).lines) {
+      return utils.setErrorStatusAndCallback(
+          cb, breakpoint, StatusMessage.BREAKPOINT_SOURCE_LOCATION,
+          utils.messages.INVALID_LINE_NUMBER + matchingScript + ':' +
+              (breakpoint.location as stackdriver.SourceLocation).line +
```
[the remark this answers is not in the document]
DominicKramer: Yep.  Fixed.

src/agent/v8/legacy-debugapi.ts
DominicKramer: Yep.  Fixed.

PR #524, opened 2018-09-11
fix: use stableObjectId field for object equality in Node 10+

### LINKED ISSUE #516: Can't dive into object properties for objects with circular references in Node 10+
Objects with circular references aren't being properly represented in Node 10+.

#524 fixes this, but problems with circular references will only disappear once a change to the V8 Inspector Protocol lands in Node 10: because the Inspector API only sends mirror objects and corresponding IDs as a result of communicating over a TCP connection (as opposed to the legacy debugging API, which runs completely in-process), there isn’t a straightforward way to detect circular references. The current behavior is that the variable table being sent through to the Stackdriver Debugger server is filled up to the brim with a breadth-first crawl of an object, or an infinite loop if `capture.maxDataSize` is set to 0.

#### Environment details

  - OS: macOS (irrelevant)
  - Node.js version: 10.8.0
  - npm version: ~6 (irrelevant)
  - `@google-cloud/debug-agent` version: 2.6.0

#### Steps to reproduce

  1. Run `test/test-circular.ts`, with the line to skip the test for Node 10+ removed.

kjin:
This allows us to properly represent objects with circular references in Node 10+, unless `capture.maxDataSize=0`, which will only be fixed once `stableObjectId` is introduced in Node 10. See #516

- [x] Tests and linter pass

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -407,7 +407,7 @@ export class Debuglet extends EventEmitter {
       // This is ignorable.
     }
 
-    if (utils.satisfies(process.version, '>=10 <10.15.2 || >=11 <11.7')) {
+    if (utils.satisfies(process.version, '>=10 <10.15.3 || >=11 <11.7')) {
       if (this.config.capture && this.config.capture.maxDataSize === 0) {
         that.logger.warn(NODE_10_CIRC_REF_MESSAGE);
       }
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -48,6 +48,10 @@ const ALLOW_EXPRESSIONS_MESSAGE = 'Expressions and conditions are not allowed' +
 const NODE_VERSION_MESSAGE =
     'Node.js version not supported. Node.js 5.2.0 and ' +
     ' versions older than 0.12 are not supported.';
+const NODE_10_CIRC_REF_MESSAGE =
+    'capture.maxDataSize=0 is not recommended on Node 10+. See ' +
+    'https://github.com/googleapis/cloud-debug-nodejs/issues/516' +
+    ' for more information.';
 const BREAKPOINT_ACTION_MESSAGE =
     'The only currently supported breakpoint actions' +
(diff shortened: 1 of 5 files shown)


kjin:
Just a note, we should not land this yet since the method for getting stable object ID will change.

kjin:
This fix should work once this is landed on v10/v11: nodejs/node#23886

DominicKramer:
@kjin Is this PR ready to land?

kjin:
@DominicKramer Not yet, see my previous comment. The PR has to get merged first, then backported to v10/v11, and then released.

kjin:
https://github.com/nodejs/node/pull/25330

kjin:
This is blocked on release of Node 10.15.2, followed by Kokoro image updates for Node 10/11

kjin:
@googleapis/node-team This is finally ready to land (after re-review), now that 10.15.3 is out. PTAL.

JustinBeckwith:
Is this really still blocked?

DominicKramer:
@JustinBeckwith I guess it is not now.  When I set the blocked label I didn't see Kelvin's last comment.

src/agent/debuglet.ts
```diff
@@ -403,6 +407,12 @@ export class Debuglet extends EventEmitter {
       // This is ignorable.
     }
 
+    if (utils.satisfies(process.version, '>=10 <10.15.3 || >=11 <11.7')) {
```
DominicKramer: This looks like it can be consolidated into a single if statement.

kjin: Done

kjin:
There will be a slight code coverage decrease because we can't pin old versions of Node to test that warning message appears on those versions. I think this is fine as we know quite clearly which change introduced the fix we needed: [10.x](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#10.15.3), [11.x](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md#11.7.0).

PR #526, opened 2018-09-17
chore: generate another key.json.enc

DominicKramer:


kinwa91:
`.gitignore` will be rewritten when synth is executed. Better remember to remove your key before committing :)

PR #535, opened 2018-09-27
fix: use well-formatted URLs when setting breakpoints

kjin:
Newer versions of Node will require that URLs be well-formatted when setting breakpoints (and will also return well-formatted URLs), as a result of https://github.com/nodejs/node/issues/22223 (this will also affect versions of Node to which that PR will be backported). Enforcing well-formatted URLs will break the Debug Agent, so this PR addresses that.

If tests pass for current versions of Node, I think this should be merged now (rather than wait for Node 11).

- [ ] Tests and linter pass


kjin:
@JustinBeckwith This PR needs work, I've been busy with other things so haven't had a chance to update it.

@soldair @DominicKramer Currently, nearly every test will fail on Node master without this change... not sure if we need to explicitly test this.

kjin:
@soldair @DominicKramer I'd appreciate a look on the newest commit. Thanks!

DominicKramer:
@ofrobots Ok will do.

DominicKramer:
This will be able to land after PR #583 lands.

Issue #559, opened 2018-10-23
Fix Node 10 test failures


DominicKramer:
The unit tests are failing for Node 10.  In particular, they are now failing for Node 10.12.0, but were passing on Node 10.1.0.

PR #560, opened 2018-10-23
 fix: fix Node 10.12.0 specific test failures

DominicKramer:
Fixes: #559


src/agent/state/inspector-state.ts
```diff
@@ -284,18 +285,26 @@ class StateResolver {
     return this.stripCurrentWorkingDirectory_(fullPath);
   }
 
+  stripFileProtocol_(path: string) {
```
ofrobots: nit: make this a static function. Ideally `isPathInNodeModulesDirectory_` would be static too but that can happen in a different PR.

PR #585, opened 2018-11-20
refactor: drop dependencies on request and lodash

JustinBeckwith:


src/agent/io/scanner.ts
```diff
@@ -83,9 +81,12 @@ class ScanResultsImpl implements ScanResults {
    *  should be included in the returned results.
    */
   selectStats(regex: RegExp): ScanStats|{} {
-    return _.pickBy(this.stats, (ignore, key) => {
-      return regex.test(key);
-    });
+    return Object.keys(this.stats).reduce((acc, key) => {
```
ofrobots: Can we use `lodash.pickby` here? If there is a reason to avoid that, extract this into a function.

JustinBeckwith:
@ofrobots ready for another look.

PR #629, opened 2019-02-08
build: test using @grpc/grpc-js in CI

kinwa91:


ofrobots:
This library doesn't even use gRPC. I would prefer if we didn't have extraneous config here.

PR #634, opened 2019-02-13
feat: make logpoint function customizable

ofrobots:
By default we use console.log for printing logpoint messages. This PR
adds a config option to allow users to customize the log function. This
can be useful in situations where the logs are being sent to 3rd parties
or need post-processing.

Fixes: https://github.com/googleapis/cloud-debug-nodejs/issues/458


ofrobots:
This is blocked on https://github.com/googleapis/cloud-debug-nodejs/pull/635 landing.

PR #635, opened 2019-02-13
chore: add missing dev dependency

DominicKramer:


ofrobots:
It might be better to review and land my deflake PR first.

DominicKramer:
I'm not able to get the system tests to fail locally using the exact same node version on either Linux or OS X.  So I'm still investigating the failure.

ofrobots:
@DominicKramer Looking at the failure, it seems like we are getting a nock failure. We seem to be trying to contact the metadata service to look up an instance property.

The e2e test already has a nock for the `projectId`, so the only remaining metadata request woul be this one:

https://github.com/googleapis/cloud-debug-nodejs/blob/1368b6217143cd175eb464779d517083d210ae84/src/agent/debuglet.ts#L386

The only reason we would reach this line is if `onGCP` is true. When you're running things locally, this would not be true. I am not sure what has changed recently for the system tests to exercise the paths differently however.

Anyhow, there are probably two ways you can fix this.
1. Add a nock for the cluster name.
2. Modify the test to provide a service context bypassing this code.

ofrobots:
I have a tentative fix for 2 in this commit you can cherry pick: https://github.com/googleapis/cloud-debug-nodejs/commit/7367af3878d071c72b9b0c34eace2a1714219cc8

Issue #616, opened 2019-01-24
Does not support webpack sourcemap


mrdulin:
#### Environment details

  - OS: macOS 10.13.6 (17G3025)
  - Node.js version: v8.11.4
  - npm version: 6.6.0
  - `@google-cloud/debug-agent` version: ^3.0.1

#### Steps to reproduce

  1. Using `webpack` build `.ts` code to `.js` and generate `.js.map` by  using webpack config

`dev-tool:  source-map`

  2. Deploy to `GAE` standard environment

When I set a breakpoint at `.ts` file, the breakpoint does not trigger.

I try using `tsc` to compiled my source code, generate `.js` and `.js.map`. After deploying, the `cloud-debug` works fine.

Here is the repo for reproduce the issue: https://github.com/mrdulin/nodejs-gcp/tree/master/src/stackdriver/debug-web-server-transpiled-code

DominicKramer:
Thanks for opening this issue and thanks for the repro.  At first glance, I don't know why the sourcemaps are not being used.  We will need to investigate and will keep this thread updated with anything we find.

soldair:
i ran your reproduction with `source-map-support/register` to see how the source maps were being written to built files.

```
node -r source-map-support/register dist/main.js
process.env.NODE_ENV: undefined
Server listening on http://localhost:8080
Error
    at module.exports.app.get (...../nodejs-gcp/src/stackdriver/debug-web-server-transpiled-code/dist/webpack:/src/main.ts:16:15)
,.....
```

it seems like the way this is being built generates sources that contain incorrect file paths. 

`dist/webpack:/src/main.ts:16:15`

DominicKramer:
I've identified that the reason why breakpoints are not hit is because when the debug agent passes Typescript file location information to the `source-map` module, that module returns that it does not know the corresponding location in the associated JavaScript file.  That is why the breakpoint does not hit.

I'm currently investigating whether the problem is in the `source-map` module itself, or if it a problem in how the debug agent is using the `source-map` API.

DominicKramer:
I have PR #640 open that will address this issue.  The problem is that the `source-map` module expects the `source` given to its API to include a `webpack` prefix because that is what is recorded in the sourcemap files generated by `webpack`.

I have a solution for this, but it requires updating the debug agent to use a newer version of `source-map`.  Since the `source-map` API changed in the new version, it will take a small amount of work to update the debug agent to use the new API.

[...]

PR #640, opened 2019-02-16
fix: add webpack support to the Sourcemapper

DominicKramer:
Fixes: #616


DominicKramer:
This PR is blocked by #643.

DominicKramer:
I was able to update this PR so that it does not depend on PR #643.

PR #663, opened 2019-03-31
test: add smoke test for app sample

JustinBeckwith:
There are currently no tests available for the samples.  This cleans up a few things and adds a simple test.  In addition removes `@google-cloud/repo-tools` and it's configuration as it wasn't being used.

### DIFF
--- a/samples/test/test.js
+++ b/samples/test/test.js
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2019, Google, LLC.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


samples/test/test.js
```diff
@@ -0,0 +1,48 @@
...
+
+'use strict';
+
+const execa = require('execa');
+
+describe('debug samples', () => {
+
+  it('should run the quickstart', done => {
+    // select a random port between 49152 and 65535
+    const PORT = Math.floor((Math.random() * (65535-49152))) + 49152;
```
bcoe: just a difference in style, I'd usually just pick a port and use it consistently for integration tests (I'd rather know that I'm not shutting down cleanly via failure to bind).

JustinBeckwith: In retrospect, that's probably going to be fine too.  Initially I was worried about the same tests running in parallel, but it doesn't matter since they're almost always in docker containers that should abstract it from us.

JustinBeckwith:
@DominicKramer from @bcoe question: 

> is there any way for us to detect that logs have been emitted to stackdriver?

ofrobots:
> is there any way for us to detect that logs have been emitted to stackdriver?

@bcoe can you elaborate/// as far as I can see there are no logs being emitted to a Stackdriver log stream. I only see `console.log`s.

bcoe:
@ofrobots I think I'm misunderstanding the default behavior of `require('@google-cloud/debug-agent').start();`, mainly I was wondering if there was a way in test to assert that it had instrumented the logging in the application -- not a blocker for me.

PR #668, opened 2019-04-04
refactor: use spawn for tests

JustinBeckwith:


samples/test/test.js
```diff
@@ -15,14 +15,16 @@
...
 
 describe('debug samples', () => {
 
   it('should run the quickstart', done => {
     // select a random port between 49152 and 65535
     const PORT = Math.floor((Math.random() * (65535-49152))) + 49152;
-    const proc = execa('node', ['app.js'], {
+    const proc = spawn('node', ['app.js'], {
+      shell: true,
       env: {
```
bcoe: I think the problem here is that `execa` merged environment by default, which is different than the default behavior.

```
env: Object.assign(process.env, {PORT})
```

👆 I think that would fix things.

JustinBeckwith: Good find!

PR #671, opened 2019-04-16
Add debugger support for Cloud Run. See b/112087600 for context

hrasadi:
(Partially) Fixes b/112087600.

- [✓] Tests and linter pass
- [✓] Code coverage does not decrease (if any source code was changed)
- [✓] Appropriate docs were updated (if necessary)

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -274,8 +274,10 @@ export class Debuglet extends EventEmitter {
     const envConfig = {
       logLevel: process.env.GCLOUD_DEBUG_LOGLEVEL,
       serviceContext: {
-        service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
-        version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+        service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME ||
+            process.env.K_SERVICE,
+        version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION ||
+            process.env.K_REVISION,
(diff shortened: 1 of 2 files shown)


src/agent/debuglet.ts
```diff
@@ -274,8 +274,10 @@ export class Debuglet extends EventEmitter {
     const envConfig = {
       logLevel: process.env.GCLOUD_DEBUG_LOGLEVEL,
       serviceContext: {
-        service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME,
-        version: process.env.GAE_VERSION || process.env.GAE_MODULE_VERSION,
+        service: process.env.GAE_SERVICE || process.env.GAE_MODULE_NAME ||
```
DominicKramer: Should `K_SERVICE` be of lowest priority or first here?  Is it possible for `GAE_MODULE_NAME` and `K_SERVICE` to both be defined but have different values?

hrasadi: By definition, GAE and Run -specific keys should not be set at the same time. These are two very separate products so the order should not really matter. I will however track this down with Cloud Run team to make sure this is the case.

DominicKramer: Sounds good.  Thanks.

DominicKramer: @hrasadi have you talked with the Cloud Run team to determine if this can land?

hrasadi: This can land. GAE and Cloud Run env vars will not clash.

DominicKramer: Great.  Thanks for looking into this.

PR #718, opened 2019-05-31
 chore: merge `options.debug` and `options`

DominicKramer:
If the `options` object given to the `start()` function has a
debug property then, previously, the options used would be
`options.debug`.  Thus anything set in `options` itself, such
as the `projectId` would not be used by the agent.  This
change suggests using the merger of `options` and `options.debug`.


src/index.ts
```diff
@@ -53,3 +52,18 @@ export function start(
 
   return agentConfig.testMode_ ? debuglet : debuglet.isReadyManager;
 }
+
+/**
+ * If the given `options` object has a `debug` property
+ * of the same type, this function returns the union of the
```
kjin: You should document which one takes precedence if both objects have identically named members, here and anywhere else where the `debug` field is referenced.

DominicKramer: Done

PR #751, opened 2019-08-20
fix: correct column numbers for line-1 breakpoints

### LINKED ISSUE #737: Line 1 breakpoint locations may no longer be accurate after 10.16+
A failing test ([testing setting line-1 breakpoints](https://github.com/googleapis/cloud-debug-nodejs/blob/bb1bcc93c2ebbd021dc193d615b7086ae5857349/test/test-v8debugapi.ts#L1822)) that is the cause of #729 highlights a potentially breaking change for us in Node 10.16: scripts are now parsed slightly differently. Consider the following code:

```
module.exports = function foo(n) { return n; }
```

In Node 10.15 and below, getting the above script via Inspector's `Debugger.getScriptSource` yields:

```
(function (exports, require, module, __filename, __dirname) { module.exports = function foo(n) { return n; }
});
```

In Node 10.16+, it yields the original source:

```
module.exports = function foo(n) { return n; }
```

This almost certainly affects line-1 breakpoints.

kjin:
Fixes #737

- [x] Tests and linter pass

### DIFF
--- a/src/agent/v8/inspector-debugapi.ts
+++ b/src/agent/v8/inspector-debugapi.ts
@@ -42,6 +42,17 @@ interface InspectorOptions {
   useWellFormattedUrl: boolean;
 }
 
+/**
+ * In older versions of Node, the script source as seen by the Inspector
+ * backend is wrapped in `require('module').wrapper`, and in new versions
+ * (Node 10.16+, Node 11.11+, Node 12+) it's not. This affects line-1
+ * breakpoints.
+ */
+const USE_MODULE_PREFIX = utils.satisfies(
+  process.version,
+  '<10.16 || >=11 <11.11'
+);
+
 export class BreakpointData {
   constructor(
     public id: inspector.Debugger.BreakpointId,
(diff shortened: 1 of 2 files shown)


src/agent/v8/inspector-debugapi.ts
```diff
@@ -42,6 +42,17 @@ interface InspectorOptions {
...
 
+/**
+ * In older versions of Node, the script source as seen by the Inspector
+ * backend is wrapped in `require('module').wrapper`, and in new versions
+ * (Node 10.16+, Node 11.11+, Node 12+) it's not. This affects line-1
+ * breakpoints.
+ */
+const USE_MODULE_PREFIX = utils.satisfies(
+  process.version,
+  '<10.16 || >=11 <11.11'
```
bcoe: I basically did the same thing in one of my libraries.

https://github.com/istanbuljs/v8-to-istanbul/blob/master/lib/v8-to-istanbul.js#L11

kjin: Thanks, I did see that `v8-to-istanbul` was mentioned in the Node PR that made this change, though I didn't realize it was your library 👍

PR #779, opened 2019-11-06
feat: Introduce javascriptFileExtensions config parameter

### LINKED ISSUE #777: Enable debugging of javascript in files with non-standard extensions
Users sometimes have javascript code in files that do not end with the standard .js extension.  The current implementation of the debugger agent does not allow these files to be debugged.

The proposal is to add a configuration setting that would allow a user to specify the file extensions that contain debuggable javascript code so that the files are correctly found and indexed.

mctavish:
This config parameter can be used to allow users to debug javascript
files that have extensions other than .js

Fixes #777 🦕

### DIFF
--- a/src/agent/config.ts
+++ b/src/agent/config.ts
@@ -180,6 +180,15 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
   appPathRelativeToRepository?: string;
 
   /**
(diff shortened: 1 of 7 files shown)


JustinBeckwith:
LGTM.  Leaving this open so @DominicKramer can take a look though

PR #792, opened 2019-11-21
test: re-enable windows tests

JustinBeckwith:
@DominicKramer we'd like to move this library to GA, but I want to make sure we clean up the issues with Windows tests first.  Mind taking a look at this and seeing what's going on?


DominicKramer:
I'm not sure offhand what is causing the problem, but I'll take a look.

PR #815, opened 2020-01-23
fix: breakpoints work on windows

DominicKramer:
On Windows, on Node 11+, the URL provided to the inspector API must start with
file:/// instead of file:// and all backslashes must be converted to forward slashes.
Otherwise, on Node 10 on windows, the URL must not start with file:// and must
contain backslashes.

### DIFF
--- a/src/agent/state/inspector-state.ts
+++ b/src/agent/state/inspector-state.ts
@@ -33,6 +33,8 @@ const GETTER_MESSAGE_INDEX = 2;
 const ARG_LOCAL_LIMIT_MESSAGE_INDEX = 3;
 
 const FILE_PROTOCOL = 'file://';
+// on windows the file protocol needs to have three slashes
+const WINDOWS_FILE_PROTOCOL = 'file:///';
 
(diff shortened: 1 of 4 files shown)


src/agent/state/inspector-state.ts
```diff
@@ -325,9 +333,18 @@ class StateResolver {
   }
 
   static stripFileProtocol_(path: string) {
-    return path.toLowerCase().startsWith(FILE_PROTOCOL)
-      ? path.substr(FILE_PROTOCOL.length)
-      : path;
+    const lowerPath = path.toLowerCase();
+    // match paths like file:///C:/... on windows
+    // but do not match paths like file:///home/... on linux
+    if (RegExp(`^${WINDOWS_FILE_PROTOCOL}[a-zA-Z]+:`).test(lowerPath)) {
```
kjin: This `RegExp` object never changes but this code path is taken often, can you make this a `static const` as well?

DominicKramer: Done

PR #851, opened 2020-03-27
build!: update to latest gts and TypeScript

AVaksman:
BREAKING CHANGE: typescript 3.7.x introduced breaking changes in the generated code.


.eslintignore
```diff
@@ -3,3 +3,6 @@ src/**/doc/*
 build/
 docs/
 protos/
+test/fixtures/fib.js
```
JustinBeckwith, line 6: Instead of calling these out individually, add `test/fixtures` and ignore the whole dir

package.json
```diff
@@ -75,7 +75,7 @@
     "cpy-cli": "^3.1.0",
     "cross-env": "^7.0.0",
     "execa": "^4.0.0",
-    "gts": "^1.0.0",
+    "gts": "^2.0.0-alpha.8",
```
JustinBeckwith, line 78: Please use `^2.0.0` :)

samples/test/test.js
```diff
@@ -14,17 +14,19 @@
 
 'use strict';
 
+// eslint-disable-next-line node/no-unpublished-require
```
JustinBeckwith, line 17: This looks like a bug in `samples/package.json`.  Can you make sure the `files` property is set in that file?  We shouldn't need this exclusion rule.

src/agent/state/legacy-state.ts
```diff
@@ -40,9 +40,11 @@ export function evaluate(
   frame: v8.FrameMirror
 ): {error: string | null; mirror?: v8.ValueMirror} {
   // First validate the expression to make sure it doesn't mutate state
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
JustinBeckwith, line 43: Instead of this, could you install `@types/acorn` and use that?

system-test/fixtures/sample/package.json
```diff
@@ -14,8 +14,8 @@
     "@google-cloud/debug-agent": "./debug.tgz"
   },
   "devDependencies": {
-    "gts": "^1.1.0",
-    "typescript": "~3.7.0",
+    "gts": "^2.0.0-alpha.8",
```
JustinBeckwith, line 17: `^2.0.0` pls

system-test/fixtures/sample/src/allowExpressionsJs.js
```diff
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// eslint-disable-next-line node/no-missing-require
 require('@google-cloud/debug-agent').start({
```
JustinBeckwith, line 15: Maybe we should add `system-test/fixtures` to the `.eslintignore`?

JustinBeckwith: (here and below)

.eslintignore
AVaksman: Added
```
test/fixtures
system-test/fixtures
```
Thanks

package.json
AVaksman: Fixed here and in `system-test`
Thanks

samples/test/test.js
AVaksman: added 
```
"files": [
    "*.js"
  ]
```
Thanks

src/agent/state/legacy-state.ts
AVaksman: used `import * as acorn from 'acorn';`
Thanks

system-test/fixtures/sample/package.json
AVaksman: fixed
Thanks

system-test/fixtures/sample/src/allowExpressionsJs.js
AVaksman: added to `. eslintignore` and removed the disabling lines
Thanks

test/fixtures/es6/transpile.js
```diff
@@ -1,7 +1,7 @@
-"use strict";
-var i = 1;
-var foo = function foo(j) {
-  return i + j + "hi";
+'use strict';
```
JustinBeckwith, line 1: If you get test failures, first place I'd look is rolling back changes to fixtures (otherwise this is fine)

AVaksman: Thanks, I was looking out for the same thing. Looks like so far everything is green.

PR #858, opened 2020-03-30
build: simplify the build

JustinBeckwith:


package.json
```diff
@@ -73,6 +72,7 @@
     "@types/tmp": "^0.1.0",
     "@types/uuid": "^7.0.0",
     "codecov": "^3.0.4",
+    "cpy-cli": "^3.1.0",
```
bcoe, line 75: I wish Node.js could just do this, without pulling in a dependency. Food for thought for the Node.js Tooling Group.

PR #929, opened 2020-10-16
build: update to latest build files

bcoe:
Ran the generator to put the library on the latest and greatest versions of all of the build files.

Sorry that I forgot to do this the other week @mctavish.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 13]
+        node: [10, 12, 14, 15]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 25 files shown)


JustinBeckwith (approved):

mctavish:
Thanks.

Unfortunately, it looks like that --recursive flag on mocha is causing some issues.  The fixtures are having build failures, and the tests under tests/misc don't run properly (and likely haven't for years).

For now, I think it makes most sense to remove the --recursive flag while fixing up those extra tests as a separate effort.

bcoe:
@mctavish we're continuing to have one failing test with the `--recursive` setting removed:

```
 1) debug-assert
       disabled
         should cover the full assert API:

      AssertionError [ERR_ASSERTION]: CallTracker does not exist on the debug assert library
      + expected - actual

      -undefined
      +function
 ```

Thoughts?

PR #951, opened 2021-04-30
feat: Add region in Debuggee labels in GCF env

### LINKED ISSUE #950: Add region in the Debuggee labels for GCF envionrment
### Background

Cloud Functions (GCF) deployments exist within the context of a cloud region. And a GCF instance can only be uniquely identified by function name + version + region. Currently the Debuggee labels contain the function name and version, but lack region, so it is not possible to identify a 1-to-1 mapping between a Debuggee and a GCF instance.

This feature request is to add the region into the Debuggee labels. It is part of the effort for better integration between Cloud Debugger and Cloud Functions.

### Implementation note
Currently the function name + version are read from the environment variables, but [the region is not available in newer GCF runtime](https://cloud.google.com/functions/docs/env-var#runtime_environment_variables_set_automatically). So the region needs to be pulled from [GCP metadata service](https://github.com/googleapis/gcp-metadata).

Louis-Ye:
* Read GCF region from either the environment or GCP metadata service
* Add the region into Debuggee labels
* Add the corresponding tests

Fixes #950

### DIFF
--- a/src/agent/debuglet.ts
+++ b/src/agent/debuglet.ts
@@ -472,6 +472,17 @@ export class Debuglet extends EventEmitter {
       that.logger.warn(NODE_10_CIRC_REF_MESSAGE);
     }
 
+    const platform = Debuglet.getPlatform();
+    let region: string | undefined;
+    if (platform === Platforms.CLOUD_FUNCTION) {
+      try {
+        region = await Debuglet.getRegion();
+      } catch (err) {
+        /* something wrong with the metadata service. */
+        that.logger.warn('Could not infer region: ', err);
+      }
+    }
+
     // We can register as a debuggee now.
     that.logger.debug('Starting debuggee, project', project);
     that.running = true;
@@ -484,9 +495,12 @@ export class Debuglet extends EventEmitter {
       sourceContext,
       onGCP,
       that.debug.packageInfo,
+      platform,
       that.config.description,
-      undefined
+      /*errorMessage=*/ undefined,
+      region
     );
+
     that.scheduleRegistration_(0 /* immediately */);
     that.emit('started');
   }
@@ -534,8 +548,10 @@ export class Debuglet extends EventEmitter {
     sourceContext: SourceContext | undefined,
     onGCP: boolean,
     packageInfo: PackageInfo,
+    platform: string,
     description?: string,
-    errorMessage?: string
+    errorMessage?: string,
+    region?: string
   ): Debuggee {
     const cwd = process.cwd();
     const mainScript = path.relative(cwd, process.argv[1]);
@@ -555,8 +571,12 @@ export class Debuglet extends EventEmitter {
       'agent.name': packageInfo.name,
       'agent.version': packageInfo.version,
       projectid: projectId,
-      platform: Debuglet.getPlatform(),
+      platform,
     };
+
+    if (region) {
+      labels.region = region;
+    }
 
     if (serviceContext) {
       if (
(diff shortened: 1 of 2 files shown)


src/agent/debuglet.ts
```diff
@@ -580,6 +600,10 @@ export class Debuglet extends EventEmitter {
       }
     }
 
+    if (region) {
+      desc += region;
```
mctavish, line 604: There probably needs to be some additional formatting here.  eg: desc += ' region: ' + region;

src/agent/debuglet.ts
```diff
@@ -637,6 +661,14 @@ export class Debuglet extends EventEmitter {
     return (await metadata.instance('attributes/cluster-name')).data as string;
   }
 
+  static async getRegion(): Promise<string> {
+    if (process.env.FUNCTION_REGION) {
+      return process.env.FUNCTION_REGION;
+    }
+    const segments = ((await metadata.instance('region')) as string).split('/');
```
mctavish, line 668: Can you include a reference to the expected format of the region string?

src/agent/debuglet.ts
```diff
@@ -637,6 +661,14 @@ export class Debuglet extends EventEmitter {
     return (await metadata.instance('attributes/cluster-name')).data as string;
   }
 
+  static async getRegion(): Promise<string> {
```
mctavish, line 664: Needs documentation, and in particular it would be good to include what specifically the returned string is expected to look like.

src/agent/debuglet.ts
mctavish, line 668: What if metadata is not available?  That is an expected use case.

src/agent/debuglet.ts
```diff
@@ -472,6 +472,17 @@ export class Debuglet extends EventEmitter {
       that.logger.warn(NODE_10_CIRC_REF_MESSAGE);
     }
 
+    const platform = Debuglet.getPlatform();
+    let region: string | undefined;
+    if (platform === Platforms.CLOUD_FUNCTION) {
+      try {
+        region = await Debuglet.getRegion();
+      } catch (err) {
+        /* something wrong with the metadata service. */
```
mctavish, line 481: Nit: the implementation of "getRegion" is bleeding out :)

mctavish (changes requested):

src/agent/debuglet.ts
Louis-Ye: I have modified the method with new comments.

src/agent/debuglet.ts
Louis-Ye: Done.

src/agent/debuglet.ts
Louis-Ye: There is no such a reference for the format of the string. I'm adding an example I saw when deployed to the real GCF environment.

src/agent/debuglet.ts
Louis-Ye: done

src/agent/debuglet.ts
Louis-Ye: I have modified the method with new comments.

src/agent/debuglet.ts
```diff
@@ -472,6 +472,12 @@ export class Debuglet extends EventEmitter {
       that.logger.warn(NODE_10_CIRC_REF_MESSAGE);
     }
 
+    const platform = Debuglet.getPlatform();
+    let region: string | undefined;
+    if (platform === Platforms.CLOUD_FUNCTION) {
+      region = await Debuglet.getRegion();
```
bcoe, line 478: rather than passing `region` into `createDebuggee`, could you simplify things and call `region = await Debuglet.getRegion();` within `createDebuggee`?

Then you could just make the `getRegion` helper return `undefined`, if `platform ~== Platforms.CLOUD_FUNCTION`.

bcoe (approved):
Small structural nit, which I think might simplify logic a bit.

Otherwise looks good to me once @mctavish approves.

mctavish (approved):

PR #957, opened 2021-05-18
fix: periodically reset v8 session to prevent memory leak

### LINKED ISSUE #811: Potential Memory Leak
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/cloud-debug-nodejs/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS: Google App Engine
  - Node.js version: 12.14.x
  - npm version: 6.13
  - `@google-cloud/debug-agent` version: 4.2.1

#### Steps to reproduce
We found that the package cause a memory leak. This image shows our application memory with the package and without it. 

image

We are running on Google App Engine with 4 nodes. Any help is appreciated. 

Thanks!

Louis-Ye:
This patch makes the cloud debugger reset the v8 debugger session to
prevent the memory leak caused by v8 breakpoint hits whenenever the
number of breakpoint hits reach the threshold.

The threshold can be set through upon the agent is started during initialization.

Fixes #811


src/agent/config.ts
```diff
@@ -357,6 +357,13 @@ export interface ResolvedDebugAgentConfig extends GoogleAuthOptions {
    * used to set a default api url
    */
   apiUrl?: string;
+
+  /**
+   * Number of times of the V8 breakpoint hits events before resetting (disable and
+   * re-enable) the breakpoints. This is to release the memory usage held by V8
+   * engine for each breakpoint hit to prevent the memory leak.
```
JustinBeckwith, line 364: Does this have a default value?  Could be nice to call out in the doc comment.

src/agent/debuglet.ts
```diff
@@ -312,6 +312,7 @@ export class Debuglet extends EventEmitter {
         minorVersion_:
           process.env.GAE_DEPLOYMENT_ID || process.env.GAE_MINOR_VERSION,
       },
+      resetV8DebuggerThreshold: process.env.RESET_V8_DEBUGGER_THRESHOLD,
```
JustinBeckwith, line 315: It looks like you may need to:
```
rm -rf node_modules package-lock.json
npm install
npm run fix
```

We had a linter update that caused problems.

Louis-Ye: Thanks for suggesting the fix. The new lint update turns out to affect lots files. I have manually included only the files that are affected by this changes.

src/agent/config.ts
Louis-Ye: Yeah there is a default value down in the bottom of this file. Will add it in the comment in next commit.

src/agent/debuglet.ts
Louis-Ye: Uh, actually, I have to include all the files to solve the lint complaint. Will do it in the next commit.

bcoe (approved):

cdbg-integration-tests (approved):

PR #964, opened 2021-06-03
fix(deps): upgrade to source-map 0.7.3

Louis-Ye:


JustinBeckwith (approved):
This is a lot cleaner!

PR #971, opened 2021-06-09
fix: surface correct error message for ambiguous sourcemap matches

### LINKED ISSUE #961: Error message is confusing when the breakpoint path matches multiple files
#### Background

This issue is discovered upon the investigation for issue #932 . Basically, the cloud debugger agent uses a fuzzy search when determining which file should a user-provided breakpoint be set on. The user will see the error message "Could not determine the output file associated with the transpiled input file" when setting a breakpoint that can potentially match to multiple files. This error message is confusing as it does not lead to a way for the user to solve the problem.

#### Environment details

  - OS: Debian 5.10.26
  - Node.js version: v12.21.0
  - npm version: 7.5.2
  - `@google-cloud/debug-agent` version: 5.2.0

#### Steps to reproduce

  1. Deploy a cloud-debugger-enabled application with multiple files having the same name but in different directories.
  2. In the GCP debugger console, upload one of the above directory as the "local source", and set a breakpoint in that same-name file.

#### Proposal to solutions

A better error message would be to show the user all the matched candidate files so that the user can make another set-breakpoint request with a more specific path.

We should either:
* show the matched candidate files in the error message directly,
* or show matched candidates in the agent logs and prompt the user to search for the logs in the error message.

Louis-Ye:
Fixes #961

### DIFF
--- a/src/agent/io/sourcemapper.ts
+++ b/src/agent/io/sourcemapper.ts
@@ -26,12 +26,25 @@ const readFilep = promisify(fs.readFile);
 
 /** @define {string} */ const MAP_EXT = '.map';
 
+/** Represents one source map file. */
 export interface MapInfoInput {
+  // The path of generated output file in the source map. For example, if
+  // "src/index1.ts" and "src/index2.ts" are generated into "dist/index.js" and
+  // "dist/index.js.map", then this field is "dist/index.js (relative to the
+  // process's working directory).
   outputFile: string;
+
+  // The source map's path (relative to the process's working directory). For
+  // the same example above, this field is "dist/index.js.map".
   mapFile: string;
+
+  // The SourceMapConsumer object after parsing the content in the mapFile.
   mapConsumer: sourceMap.SourceMapConsumer;
-  // the sources are in ascending order from
-  // shortest to longest
+
+  // The original sources in the source map. Each value is relative to the
+  // source map file. For the same example above, this field is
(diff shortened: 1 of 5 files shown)


test/test-sourcemapper.ts
```diff
@@ -69,7 +69,7 @@ function testTool(
         ' it states that it has mapping info for files it knows about',
       done => {
         assert.strictEqual(
-          sourcemapper.hasMappingInfo(inputFilePath),
+          sourcemapper.getMapInfoInput(inputFilePath) !== null,
```
JustinBeckwith, line 72: nit, we could use `assert.notStrictEqual(sourcemapper.getMapInfoInput(inputFilePath), null, ...)`

JustinBeckwith (approved):

test/test-sourcemapper.ts
Louis-Ye: Done.

src/agent/io/sourcemapper.ts
```diff
@@ -26,12 +26,25 @@ const readFilep = promisify(fs.readFile);
 
 /** @define {string} */ const MAP_EXT = '.map';
 
+/** Represents one source map file. */
```
mctavish, line 29: Thank you so much for this documentation!

mctavish (approved):

PR #977, opened 2021-06-16
fix: Add debugging information for sourcemapper

### LINKED ISSUE #932: Could not determine the output file associated with the transpiled input file
### Problem you have encountered:
Receiving "Could not determine the output file associated with the transpiled input file" error when we deploy node js code.

### Configurations:
- typescript - v4.0.3
- node - v14.15.0
- webpack - v5.4.0

We tried to enable sourcemap with below webpack sourcemapers

- 'source-map'
- 'inline-source-map'
- 'hidden-source-map'
But we are not able to set debugger in typescript based project.

### What you expected to happen:
The debugger should be attached to a node js based project which contains code in TS and transpile through webpack.

Louis-Ye:
This PR is to add debugging information to source mapper so that it is easier to debug related issues from users.

Fixes #932

### DIFF
--- a/test/test-sourcemapper.ts
+++ b/test/test-sourcemapper.ts
@@ -57,25 +57,34 @@ function testTool(
     const logger = new MockLogger();
     let sourcemapper: sm.SourceMapper;
 
-    it('for tool ' + tool + ' sourcemapper should be created correctly',
-       async () => {
-         const start = Date.now();
-         sourcemapper = await sm.create([mapFilePath], logger);
-         assert(
-             Date.now() - start < QUICK_MILLISECONDS,
-             'should create the SourceMapper quickly');
+    it(
+      'for tool ' + tool + ' sourcemapper should be created correctly',
+      async () => {
+        const start = Date.now();
+        sourcemapper = await sm.create([mapFilePath], logger);
+        assert(
+          Date.now() - start < QUICK_MILLISECONDS,
+          'should create the SourceMapper quickly'
+        );
 
-         // Verify if the debugging information is correctly printed.
-         assert.notStrictEqual(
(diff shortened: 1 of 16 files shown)


test/test-sourcemapper.ts
```diff
@@ -52,16 +54,37 @@ function testTool(
...
+      async () => {
+        const start = Date.now();
+        sourcemapper = await sm.create([mapFilePath], logger);
+        assert(
+          Date.now() - start < QUICK_MILLISECONDS,
+          'should create the SourceMapper quickly'
+        );
+
+        // Verify if the debugging information is correctly printed.
+        assert.notStrictEqual(
```
mctavish, line 71: Could be just me, but I find the notStrictEqual(...indexOf(...), -1) to be hard to read.  Maybe matches instead?

test/test-sourcemapper.ts
```diff
@@ -52,16 +54,37 @@ function testTool(
...
+      'for tool ' + tool + ' sourcemapper should be created correctly',
+      async () => {
+        const start = Date.now();
+        sourcemapper = await sm.create([mapFilePath], logger);
+        assert(
+          Date.now() - start < QUICK_MILLISECONDS,
+          'should create the SourceMapper quickly'
+        );
+
+        // Verify if the debugging information is correctly printed.
```
mctavish, line 70: Split this into a different test, and please provide an example of what the output is.  It's hard for me to parse it based off of the assertions below.

test/test-sourcemapper.ts
Louis-Ye: Done.

test/test-sourcemapper.ts
Louis-Ye: Done.

mctavish (approved):

PR #978, opened 2021-06-16
fix: source mapping original path instead of user-provided input

### LINKED ISSUE #963: Breakpoint is set to a wrong line when the user-provided breakpoint path is a partial path
#### Background

This is a bug discovered during the investigation for #932. When the user-provided breakpoint request has the path that is not a full relative path to the root directory of the application, (for example `index.ts` instead of `src/index.ts`), the breakpoint will be set to a wrong line (line 0) because the relative path of the target file to the source map is wrongly generated.

#### Environment details

  - OS: Debian 5.10.26
  - Node.js version: v12.21.0
  - npm version: 7.5.2
  - `@google-cloud/debug-agent` version: 5.2.0

Louis-Ye:
Uses the original source path in the mapInfo instead of user-provided input to check the generated sources, so that when the user provided a partial match of the file, we can still correctly map the original file to the generated file.

This PR is based on PR #977 

Fixes #963

### DIFF
--- a/src/agent/io/sourcemapper.ts
+++ b/src/agent/io/sourcemapper.ts
@@ -35,8 +35,9 @@ export interface MapInfoInput {
   // process's working directory).
   outputFile: string;
 
-  // The path of the original source file. In the example above, this will be
-  // either "src/index1.ts" or "src/index2.ts".
+  // The path of the original source file (relative to the process's working
+  // directory). In the example above, this will be either "src/index1.ts" or
+  // "src/index2.ts".
   inputFile: string;
 
   // The source map's path (relative to the process's working directory). For
--- a/src/agent/io/sourcemapper.ts
+++ b/src/agent/io/sourcemapper.ts
@@ -230,7 +230,7 @@ export class SourceMapper {
    * @param {number} (Optional) The column number in the line of the file
    *   specified where the column number is zero-based.
    * @param {string} The entry of the source map info in the sourceMapper. Such
-   *     an entry is supposed to get got by the getMapInfoInput method.
+   *     an entry is supposed to be got by the getMapInfoInput method.
    *
    * @return {Object} The object returned has a "file" attribute for the
    *   path of the output file associated with the given input file (where the
--- a/src/agent/io/sourcemapper.ts
+++ b/src/agent/io/sourcemapper.ts
@@ -327,7 +327,10 @@ export class SourceMapper {
  * @param {Logger} logger A logger that reports errors that occurred while
  *  processing the given sourcemap files
  */
-export async function create(sourcemapPaths: string[], logger: Logger): Promise<SourceMapper> {
+export async function create(
+  sourcemapPaths: string[],
+  logger: Logger
+): Promise<SourceMapper> {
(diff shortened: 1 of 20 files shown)


src/agent/io/sourcemapper.ts
```diff
@@ -167,17 +174,10 @@ async function processSourcemap(
...
-  /**
-   * @param {Array.<string>} sourcemapPaths An array of paths to .map sourcemap
-   *  files that should be processed.  The paths should be relative to the
-   *  current process's current working directory
-   * @param {Logger} logger A logger that reports errors that occurred while
-   *  processing the given sourcemap files
-   * @constructor
-   */
-  constructor() {
+  constructor(readonly logger: Logger) {
```
mctavish, line 180: Private constructor?

src/agent/io/sourcemapper.ts
```diff
@@ -34,6 +35,10 @@ export interface MapInfoInput {
   // process's working directory).
   outputFile: string;
 
+  // The path of the original source file. In the example above, this will be
```
mctavish, line 38: Please specify what this path is relative to.

test/test-sourcemapper.ts
```diff
@@ -234,11 +267,19 @@ testTool(
   ]
 );
 
+testTool(
```
mctavish, line 270: Is this the test that repros the original issue?  If so, please add some additional description or pointer to the bug because I'm a bit too dense to figure out what this is verifying.

JustinBeckwith (approved):
LGTM from a TypeScript perspective

test/test-sourcemapper.ts
Louis-Ye: Added the comments.

src/agent/io/sourcemapper.ts
Louis-Ye: Done.

src/agent/io/sourcemapper.ts
Louis-Ye: The create() function is actually not a static function of this class, marking the constructor as private will change the usage of this function in all places. I think for this PR, it is better to keep it as it is, what do you think?

src/agent/io/sourcemapper.ts
```diff
@@ -218,13 +225,12 @@ export class SourceMapper {
...
 
   /**
-   * @param {string} inputPath The path to an input file that could possibly
-   *  be the input to a transpilation process.  The path should be relative to
-   *  the process's current working directory
    * @param {number} The line number in the input file where the line number is
    *   zero-based.
    * @param {number} (Optional) The column number in the line of the file
    *   specified where the column number is zero-based.
+   * @param {string} The entry of the source map info in this.infoMap.
```
mctavish, line 232: Please update this comment; the type isn't a string, and I'm not able to understand what this parameter should be based on this documentation.

mctavish (approved):
One small change to make, but looks good.

src/agent/io/sourcemapper.ts
Louis-Ye: Done.

PR #1067, opened 2022-04-22
feat: Add ability to store breakpoint data in firebase realtime database

### LINKED ISSUE #1071: Add ability to store breakpoint data in firebase realtime database
The firebase realtime database has similar functionality to the debugger backend.  Allow users to connect to the firebase realtime database as a backend in order to more directly control their data.

mctavish:
This is a work in progress:
- Firebase-related code may need updates to hit desired code quality.
- Firebase-related code does not yet have unit tests
- Some OnePlatform-related tests have been temporarily disabled.
- Documentation has not been updated

The intention is to get reviews going on this code as development continues on this branch.  Once the above issues (and any others that come up in review) are addressed, the branch can be pulled into main.

----

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/cloud-debug-nodejs/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

Fixes #<issue_number_goes_here> 🦕


src/agent/config.ts
```diff
@@ -412,4 +427,6 @@ export const defaultConfig: ResolvedDebugAgentConfig = {
   forceNewAgent_: false,
   testMode_: false,
   resetV8DebuggerThreshold: 30,
+
+  useFirebase: false,
```
bcoe, line 431: If we were ever to make `firebase` the default, would we remove this toggle?

src/agent/debuglet.ts
```diff
@@ -35,11 +35,13 @@ import {
   ResolvedDebugAgentConfig,
 } from './config';
 import {Controller} from './controller';
+import {OnePlatformController} from './oneplatformcontroller';
```
bcoe, line 38: I like this abstraction 👍

src/agent/debuglet.ts
```diff
@@ -426,21 +435,26 @@ export class Debuglet extends EventEmitter {
...
-      that.logger.error(
-        'The project ID could not be determined: ' + err.message
-      );
-      that.emit('initError', err);
-      return;
+    if (this.config.useFirebase) {
+      // Project id was discovered during initialization of the app.
+      project = (this.controller as FirebaseController).getProjectId();
+    } else {
+      try {
```
bcoe, line 442: reading the code, I don't see any reason to create a reference `that = this`, as long as you use closures of the form:

```js
() => {
 ...
}
```

`this` will reference the top level scope. I think this is probably hold over that's worth cleaning up.

src/agent/debuglet.ts
```diff
@@ -559,13 +570,14 @@ export class Debuglet extends EventEmitter {
...
+    // Incompatibilities!  Keys must be non-empty strings and can't contain ".", "#", "$", "/", "[", or "]"
     const labels: {[key: string]: string} = {
       'main script': mainScript,
-      'process.title': process.title,
+      'process title': process.title,
       'node version': process.versions.node,
       'V8 version': process.versions.v8,
-      'agent.name': packageInfo.name,
-      'agent.version': packageInfo.version,
+      'agent name': packageInfo.name,
```
bcoe, line 579: We should probably call this a breaking change, I'm assuming the issue is that firebase can't handle a `.` in keynames?

Perhaps we could use a `_` rather than a ` `, that reads a bit better to me and is easier to search.

src/agent/debuglet.ts
```diff
@@ -748,129 +765,36 @@ export class Debuglet extends EventEmitter {
             }
           ).debuggee.id;
           // TODO: Handle the case when `result` is undefined.
-          that.emit('registered', (result as {debuggee: Debuggee}).debuggee.id);
-          that.debuggeeRegistered.resolve();
+          that.emit('registered', (result as {debuggee: Debuggee}).debuggee.id); // FIXME: Do we need this?
```
bcoe, line 768: I would make sure to open tracking issues for these FIXME comments.

src/agent/firebasecontroller.ts
```diff
@@ -0,0 +1,224 @@
...
+    let credential = undefined;
+    let projectId = undefined;
+    if (options.keyPath) {
+      // Use the project id and credentials in the path specified by the keyPath.
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const serviceAccount = require(options.keyPath);
+      projectId = serviceAccount['project_id'];
+      credential = firebase.credential.cert(serviceAccount);
+    } else {
+      // TODO: Find out how to find the project ID.
```
bcoe, line 50: I would create an issue for this too.

src/agent/firebasecontroller.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    // Indicate success to the caller.
+    callback();
+  }
+
+  subscribeToBreakpoints(
+    debuggee: Debuggee,
+    callback: (err: Error | null, breakpoints: stackdriver.Breakpoint[]) => void
+  ): void {
+    console.log('Started subscription for breakpoint updates');
```
bcoe, line 199: I would use https://nodejs.org/api/util.html#utildebuglogsection-callback rather than `console.log`, that way you don't have to remember to remove all of the debug lines from your code, and if you make the debug lines descriptive, it's even worth leaving them in.

src/agent/oneplatformcontroller.ts
```diff
@@ -0,0 +1,332 @@
+// Copyright 2022 Google LLC
```
bcoe, line 1: nit: I would probably use hyphens for these file names:

```
oneplatform-controller, firebase-controller
```

bcoe (changes requested):
Left mostly nits, this code reads fairly clean to me.

src/agent/config.ts
mctavish: Yes we would, and in the process would also retire a number of config values that would no longer apply.

system-test/test-controller.ts
```diff
@@ -85,7 +86,7 @@ describe('Controller', function () {
 
   it('should pass success on timeout', done => {
```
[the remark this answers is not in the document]
mctavish: Agreed!  The challenge that I have at the moment is finding a suitable mock for the database.  nock doesn't do the trick since the client isn't using the rest api.  Any suggestions?

src/agent/debuglet.ts
mctavish: Correct; and I like the underscore idea.

This shouldn't be a breaking change (at least for users of the agent) but given the addition of the configuration and new dependencies this feature could be considered to be significant enough to warrant a major version bump.  Thoughts?

src/agent/oneplatformcontroller.ts
mctavish: I like it.  Done.

package.json
```diff
@@ -50,6 +50,7 @@
     "console-log-level": "^1.4.0",
     "extend": "^3.0.2",
     "findit2": "^2.2.3",
+    "firebase-admin": "^9.11.1",
```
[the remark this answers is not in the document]
mctavish: We're pulling in the app, credentials, and database portions of the dependency, and there don't appear to be more focused distributions of these libraries.  The firebase-admin package is much smaller than the regular firebase one though, so hopefully the impact will be less.

There aren't any ways to hint to node that we only care about those three parts of the package, are there?

src/agent/firebasecontroller.ts
mctavish: Done.  Thanks.

src/agent/debuglet.ts
mctavish: Thanks! :)

test/test-debuglet.ts
```diff
@@ -1169,7 +1169,8 @@ describe('Debuglet', () => {
       debuglet.start();
     });
 
-    it('should have breakpoints fetched when promise is resolved', function (done) {
+    /*it('should have breakpoints fetched when promise is resolved', function (done) {
```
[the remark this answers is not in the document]
mctavish: Created https://github.com/googleapis/cloud-debug-nodejs/issues/1072 to specifically track this

src/agent/debuglet.ts
bcoe: Given all the restructuring, I might call it a breaking change, even if there's no one specific breakage we're calling attention to.

bcoe (approved):

PR #1077, opened 2022-05-11
fix: use debuglet's logger in the controller.

mctavish:
This change will ensure consistent logging behaviour across the codebase.


jasonborg (approved):
My only comment is I was expecting to see a similar change to pass the logger into the FirebaseController, but I see it uses 'const debuglog = util.debuglog('cdbg.firebase');'

We're ok with this for now?

mctavish:
Okay for now.  The logs in the firebase controller are explicitly debug
logs that Ben suggested I use instead of console.log during development.
We'll revisit what logging should be present later.

PR #1097, opened 2022-07-25
feat: improve debuggee ids

### LINKED ISSUE #1096: Switch firebase-backend debuggee id formats to d-********
Debuggee ids generated by the firebase client in this package are 32 character hex strings.  This seems excessive when avoiding collisions within a project and makes it more difficult to use the CLI.

Switching the debuggee ids to a format such as "d-{8-digit-hex}" will make it shorter and the "d-" will add a small amount of context for the seemingly random characters.

mctavish:
Debuggee ids are now of the form: d-********

Fixes #1096 🦕

### DIFF
--- a/src/agent/firebase-controller.ts
+++ b/src/agent/firebase-controller.ts
@@ -133,10 +133,11 @@ export class FirebaseController implements Controller {
 
     // Calculate the debuggee id as the hash of the object.
     // This MUST be consistent across all debuggee instances.
-    this.debuggeeId = crypto
+    const debuggeeHash = crypto
       .createHash('md5')
       .update(JSON.stringify(debuggee))
       .digest('hex');
+    this.debuggeeId = `d-${debuggeeHash.substring(0, 8)}`;
     debuggee.id = this.debuggeeId;
 
     const debuggeeRef = this.db.ref(`cdbg/debuggees/${this.debuggeeId}`);
--- a/src/agent/firebase-controller.ts
+++ b/src/agent/firebase-controller.ts
@@ -133,8 +133,10 @@ export class FirebaseController implements Controller {
 
     // Calculate the debuggee id as the hash of the object.
     // This MUST be consistent across all debuggee instances.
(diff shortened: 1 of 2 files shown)


src/agent/firebase-controller.ts
```diff
@@ -133,10 +133,11 @@ export class FirebaseController implements Controller {
 
     // Calculate the debuggee id as the hash of the object.
     // This MUST be consistent across all debuggee instances.
-    this.debuggeeId = crypto
+    const debuggeeHash = crypto
       .createHash('md5')
       .update(JSON.stringify(debuggee))
```
jasonborg, line 138: Unrelated to this change, but something to note, it seems JSON.stringify may not be deterministic in the ordering of keys: https://github.com/auth0/node-jsonwebtoken/issues/404. It's possible in practice the way we end up using things and creating the debuggee it may not be an issue for us.

src/agent/firebase-controller.ts
```diff
@@ -133,10 +133,11 @@ export class FirebaseController implements Controller {
 
     // Calculate the debuggee id as the hash of the object.
     // This MUST be consistent across all debuggee instances.
-    this.debuggeeId = crypto
+    const debuggeeHash = crypto
       .createHash('md5')
```
jasonborg, line 137: Just realized we're using MD5 here, but SHA1 in the Python and Java agents.

jasonborg (approved):

src/agent/firebase-controller.ts
mctavish: Good catch.  I'll switch to SHA1.  MD5 isn't FIPS compliant so it's possible it won't be present.

src/agent/firebase-controller.ts
mctavish: I'll leave this as a TODO for now, but you're right.  The order appears to be based on insertion order which *should* be deterministic for these debuggee objects, but it would be good to be able to guarantee that it's stable.

PR #1116, opened 2022-12-01
feat: add error handling to firebase controller.

### LINKED ISSUE #1115: Add error handling to the firebase controller
The firebase controller currently lacks sufficient error handling.  While the rest of the agent code has retry logic in place to handle failed registrations, breakpoint updates, etc, the firebase controller swallows those errors and does not surface them properly.

mctavish:

Fixes #1115 🦕

### DIFF
--- a/src/agent/firebase-controller.ts
+++ b/src/agent/firebase-controller.ts
@@ -74,25 +74,41 @@ export class FirebaseController implements Controller {
     if (options.databaseUrl) {
       databaseUrl = options.databaseUrl;
     } else {
-      // TODO: Test whether this exists.  If not, fall back to -default.
+      // TODO: Add fallback to -default
       databaseUrl = `https://${projectId}-cdbg.firebaseio.com`;
     }
 
+    let app : firebase.app.App;
     if (credential) {
-      firebase.initializeApp({
+      app = firebase.initializeApp({
         credential: credential,
         databaseURL: databaseUrl,
       }, 'cdbg');
     } else {
       // Use the default credentials.
-      firebase.initializeApp({
+      app = firebase.initializeApp({
         databaseURL: databaseUrl,
       }, 'cdbg');
     }
 
     const db = firebase.database();
 
-    // TODO: Test this setup and emit a reasonable error.
+    // Test the connection by reading the schema version.
+    try {
+      const version_snapshot = await db.ref('cdbg/schema_version').get();
+      if (version_snapshot) {
+        const version = version_snapshot.val();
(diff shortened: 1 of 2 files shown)


src/agent/firebase-controller.ts
```diff
@@ -74,25 +74,49 @@ export class FirebaseController implements Controller {
...
+      } else {
+        app.delete();
+        throw new Error('failed to fetch schema version from database');
+      }
+    } catch (e) {
+      app.delete();
+      throw e;
+    }
+
     debuglog('Firebase app initialized.  Connected to', databaseUrl);
```
jasonborg, line 120: This log is now a duplicate, there is now line 107 above.

src/agent/firebase-controller.ts
```diff
@@ -224,26 +263,53 @@ export class FirebaseController implements Controller {
...
+    );
   }
 
   stop(): void {
     if (this.bpRef) {
       this.bpRef.off();
       this.bpRef = undefined;
     }
+    try {
+      firebase.app('cdbg').delete();
```
jasonborg, line 310: May be worth either storing the returned app from the initialize, or using a defined constant for the app name, 'cdbg' to ensure it's used the same in the initialize and this cleanup.

src/agent/firebase-controller.ts
```diff
@@ -184,32 +212,43 @@ export class FirebaseController implements Controller {
     // https://firebase.google.com/docs/reference/rest/database#section-server-values
     breakpoint_map['finalTimeUnixMsec'] = {'.sv': 'timestamp'};
 
-    this.db
-      .ref(`cdbg/breakpoints/${this.debuggeeId}/active/${breakpoint.id}`)
-      .remove();
+    try {
+      await this.db
+        .ref(`cdbg/breakpoints/${this.debuggeeId}/active/${breakpoint.id}`)
+        .remove();
```
jasonborg, line 218: Updates get queued up and retried on failure outside of here right? Will if remove() has already been done successfully on one attempt, I'm assuming it won't 'fail' on a subsequent?

jasonborg (approved):

src/agent/firebase-controller.ts
mctavish: I've confirmed that removing something that is not present will succeed.

Yes, updates are retried if they fail.

src/agent/firebase-controller.ts
mctavish: Good catch; removed.

src/agent/firebase-controller.ts
mctavish: Using a constant.

PR #1121, opened 2022-12-08
feat: add active debuggee support

### LINKED ISSUE #1120: Add active debuggee support for the firebase backend
The snapshot debugger CLI would like to be able to list debuggees in order of most recent activity: 
 https://github.com/GoogleCloudPlatform/snapshot-debugger/issues/76

In order to support this, the debug agent will need to set the registration time and periodically update a second timestamp in the database.

mctavish:

Fixes #1120 🦕

### DIFF
--- a/src/agent/firebase-controller.ts
+++ b/src/agent/firebase-controller.ts
@@ -37,7 +37,7 @@ export class FirebaseController implements Controller {
   bpRef?: firebase.database.Reference;
 
   markActiveInterval: ReturnType<typeof setInterval> | undefined;
-  markActivePeriod: number = 60 * 60 * 1000; // 1 hour in ms.
+  markActivePeriodMsec: number = 60 * 60 * 1000; // 1 hour in ms.
 
   /**
    * Connects to the Firebase database.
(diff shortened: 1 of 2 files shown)


src/agent/firebase-controller.ts
```diff
@@ -36,6 +36,9 @@ export class FirebaseController implements Controller {
   debuggeeId?: string;
   bpRef?: firebase.database.Reference;
 
+  markActiveInterval: ReturnType<typeof setInterval> | undefined;
+  markActivePeriod: number = 60 * 60 * 1000; // 1 hour in ms.
```
jasonborg, line 40: Consider changing the variable name to `markActivePeriodMsec`, then there's no need to find the definition to find the comment indicating the units.

src/agent/firebase-controller.ts
```diff
@@ -300,6 +320,26 @@ export class FirebaseController implements Controller {
         callback(e, []);
       }
     );
+
+    this.startMarkingDebuggeeActive();
+  }
+
+  startMarkingDebuggeeActive() {
+    console.log(`starting to mark every ${this.markActivePeriod} ms`);
+    this.markActiveInterval = setInterval(() => {
```
jasonborg, line 329: Rename to `markActiveIntervalId` ? it wasn't until I read the setInterval documentation that I realized it returns an ID.

jasonborg (approved):

src/agent/firebase-controller.ts
jasonborg: Ah, interesting, and fair enough.

PR #1127, opened 2023-01-09
docs: Update Firebase notes in README

jasonborg:


mctavish:
You'll need to change .readme-partials.yaml instead of the readme directly.

Issue #1143, opened 2023-05-31
Remove support for deprecated Cloud Debugger API


mctavish:
The Cloud Debugger API is deprecated and will be shut down at the end of May 2023.

This feature request is for the agent to remove support for that API and to have all agents use the Firebase backend instead.

PR #1149, opened 2023-06-06
docs: remove content about the shut down Cloud Debugger service

mctavish:
This PR updates the README content to reflect the changes made for #1143


jasonborg (approved):
Some other parts of the readme could also use updating if you're able to, such as references to the `Cloud Debugger` and the `Before you begin` section that says to `Enable the Cloud Debugger API1

PR #1192, opened 2024-01-30
deps: Update p-limit to ^4.0.0

mctavish:
Due to node version upgrades, we've missed the dependabot updates for this package.  Catching up a bit.


mctavish:
Oof.  Abandoning this; the compiled typescript is creating a "require" for p-limit which isn't going to work.  Not worth the effort to track this down.

